PR #1, opened 2018-09-26
Changes made to enable the reading of native msg data file in SIFT

ColinDuff:
Here are the changes needed to enable us to read the native MSG data using SIFT.

NB you may want to cherry pick the updates in importer.py as the problems with the start and end time will be fixed with a satpy update.

The missing attribute problems eg sat.name m inst.name may also be an implementation issue as i wasnt sure what updates are need in common.py etc


djhoese:
Closed by 9dc60ccf31bf4362b71da380cb50a1c10e436867

Still requires updates from SatPy to work for msg reader. Coming soon hopefully...

djhoese:
NOTE: This does not fix the RGB issues you have seen. That's something we'll need to look in to.

PR #11, opened 2018-11-16
Add sqrt status to saved colormap

rdaruwala:
SQRT status now saves on SIFT closing / reopening


sift/ui/GradientControl.py
```diff
@@ -120,7 +122,7 @@ def cloneGradient(self):
 
                 self.user_colormap_states[save_name] = self.ColorBar.saveState()
             self.updateListWidget(save_name)
-            self.saveNewMap(self.ColorBar.saveState(), save_name)
+            self.saveNewMap(self.user_colormap_states[save_name], save_name)
```
djhoese: I know this isn't your change, but does this method re-save the colormap even if the person says "cancel"?

PR #16, opened 2019-02-21
Fix presentation for newly layers not matching similar layers

djhoese:
@jgerth pointed out that doing the following results in the new layers not having the right presentation:

1. Load B01 data
2. Change the B01 colormap/gamma/color limits
3. Load more B01 data

This PR fixes that.


djhoese:
You mean like:

```
colormap=cmap if family_prez is None else family_prez.colormap
```

djhoese:
Closes #209

Issue #216, opened 2019-03-04
Refactor names and modules to be more PEP8 compliant


djhoese:
@rayg-ssec and I, in person, have discussed renaming certain variables and modules in SIFT to be more PEP8 compliant. Key changes being that variables are lowercase underscored, classes are camelCase, modules are lowercase underscored, etc. I'd also like to take this opportunity to remove modules or argparse boilerplate that are no longer used. Ray and I plan to perform this work on Thursday, but I thought I'd start a list:

- [x] `sift.common.prez` -> `Prez`
- [x] `sift.common.span` -> `Span`
- [x] `sift.view.*` need to be lowercased
- [x] `sift.ui.GradientControl` lowercased
- [x] `sift.model.colormap` remove
- [x] `sift.common.*` [Enums](https://docs.python.org/3/library/enum.html#creating-an-enum) should be named like regular classes with all caps members
- [x] `flake8 sift` and fix all issues

Feel free to comment for anything else that should be added to the list.

djhoese:
Docstring convention: Google style (need to add in READMEs and dev docs): https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

djhoese:
Closed with #217

PR #217, opened 2019-03-07
Refactor code to be more PEP8 compliant

djhoese:
See #216 for more details.


djhoese:
`flake8 sift` now passes without error

PR #221, opened 2019-03-11
Refactor MainWindow initialization and control logic into smaller building blocks

rayg-ssec:
Refactoring of MainWindow to
- separate __init__ into smaller routines with better descriptions (it was ridiculously long)
- move animation-specific use cases to UserControlsAnimation class.

Expect a follow-up to this "training-wheels" PR with probe behavior similarly separated from (and composed back into) MainWindow.


sift/__main__.py
```diff
@@ -226,12 +226,202 @@ def _common_path_prefix(paths):
         return None
 
 
-class Main(QtGui.QMainWindow):
-    _last_open_dir = None  # directory to open files in
-    _recent_files_menu = None  # QMenu
+class UserControlsAnimation(QtCore.QObject):
+    """Controller behavior focused around animation bar and next/last key bindings.
+    Connects between document, scene graph manager, layer list view of document; uses UI elements.
```
djhoese: Could you add a blank line between the first "subject line" and the body of the docstring?

sift/__main__.py
```diff
@@ -226,12 +226,202 @@ def _common_path_prefix(paths):
...
+        self.ui.animationSlider.repaint()
+        if animating:
+            self.ui.animationLabel.setText(self.document.time_label_for_uuid(uuid))
+        else:
+            self.update_frame_time_to_top_visible()
+
+    def update_frame_time_to_top_visible(self, *args):
+        """Update frame slider's time display to show the current top layer's time."""
+        # FUTURE: don't address layer set directly
+        self.ui.animationLabel.setText(self.document.time_label_for_uuid(self.scene_manager.layer_set.top_layer_uuid()))
```
djhoese: Doesn't the doc have a method/property for what should be the current top layer?

sift/__main__.py
```diff
@@ -653,33 +707,15 @@ def __init__(self, config_dir=None, workspace_dir=None, cache_size=None, glob_pa
...
-
-        # allow animation slider to set animation frame being displayed:
-        self.ui.animationSlider.valueChanged.connect(self.animation_slider_jump_frame)
+        self._init_layer_panes()
+        self._init_rgb_pane()
+        self._init_map_widget()
 
-        # allow animation, once stopped, to propagate visibility to the document and layerlist:
-        self.scene_manager.didChangeLayerVisibility.connect(self.document.animation_changed_visibility)
+        self.animation = UserControlsAnimation(self.ui,
```
djhoese: Is `animation` too ambiguous?

rayg-ssec: I think it's okay, though animation_behavior might be appropriate - but would that turn into more questions?

sift/__main__.py
rayg-ssec: document.current_visible_layer_uuid may be appropriate if it's equivalent...

sift/__main__.py
rayg-ssec: done

sift/__main__.py
djhoese: Possibly. Only other thought is if we want to have a naming convention for behaviors. I'm ok leaving it.

sift/__main__.py
djhoese: I just approved this but realized this was never changed. Are we going to regret not changing this in the future?

rayg-ssec: Just pushed a one-liner to use the preferred document property; quick testing shows things still working as expected.

sift/__main__.py
```diff
@@ -226,12 +227,203 @@ def _common_path_prefix(paths):
...
+        self.ui.animationSlider.repaint()
+        if animating:
+            self.ui.animationLabel.setText(self.document.time_label_for_uuid(uuid))
+        else:
+            self.update_frame_time_to_top_visible()
+
+    def update_frame_time_to_top_visible(self, *args):
+        """Update frame slider's time display to show the current top layer's time."""
+        self.ui.animationLabel.setText(self.document.time_label_for_uuid(self.document.current_visible_layer_uuid))
+        # self.ui.animationLabel.setText(self.document.time_label_for_uuid(self.scene_manager.layer_set.top_layer_uuid()))
```
djhoese: Remove the commented out line?

rayg-ssec: Done.

PR #222, opened 2019-03-12
Migrate to PyQt5

### LINKED ISSUE #220: Migrate to PyQt5
SIFT currently uses PyQt4. PyQt5 has been around for a long time and PyQt4 is basically deprecated from conda/conda-forge. So we need to update.

In the past we had tried to migrate to PyQt5 but there were bugs due to vispy using the old PyQt5 GL API and us using a web view for the colormap display. It resulted in weird transparency in the GL canvas and random black boxes rendering over various widgets.

As part of this, I will probably move to python 3.7 as the default build. After SIFT 1.1 gets released we may want to remove the custom versions of libraries that we include on the `larch` conda channel (pyqt, matplotlib, etc).

djhoese:
Close #220.

Use https://github.com/rferrazz/pyqt4topyqt5 to convert PyQt4 code to PyQt5. Had to clean up a few things where the tool tried to `from PyQt5.QtWidgets import *` instead of importing the correct classes.

Also had to comment out the timeline creation because it currently uses the legacy GL API (QGLWidget) which causes the same black block issues.


djhoese:
The WIP state of this PR is completely dependent on the timeline stuff. @rayg-ssec if you come up with a solution then I'm ok merging this as long as all devs understand the environment changes this requires.

rayg-ssec:
Given that I don't have a PyQt5/py37 test environment yet, I tested removal of QGLWidget from the timeline view (it acts as a performance optimization when working with complex QGraphicsScenes). See rayg-ssec/sift:refactor-remove-qglwidget for the patch. 

The extent of the rest of the changes seem pretty mechanical (aside from _fromUtf8 here and there), so if the timeline patch allows it to operate adequately in an interim solution, I'm okay with merging this.

djhoese:
@rayg-ssec I merged your branch in to this branch and pushed it here. I think this is ready to merge then. I need to make vispy builds for Windows and Linux platforms still, but that only matters for developers. Will finish it today/now.

djhoese:
I've added the new versions of vispy to the larch conda channel for python 3.6 and python 3.7. You *should* (assuming you haven't set strict channel priority) be able to install vispy and quamash for a new environment with PyQt5. This *should* be less of a headache than using PyQt4.

Any visual issues with using PyQt5 will be addressed in a different PR.

PR #230, opened 2019-04-04
Rename package name for conda-forge and PyPI

### LINKED ISSUE #227: Rename package name for conda-forge and PyPI
This project's name and package name is "sift". This in general is not a problem, but as we expand to other build/package tools and hosting services we may have collisions with other existing software projects. For example on PyPI there is already a "sift" package: https://pypi.org/project/Sift/. If we want people to be able to `pip install sift` we will have to pick a different name. There is currently no package named "sift" on conda-forge.

If would be nice in the future if people could `pip install sift` or `conda install sift` without having to specify a special UW-specific channel. Additionally it would be nice if we could take advantage of conda-forge's build bots and environments. I'd like to propose that we rename the name of the package to something like "uwsift" or "ssecsift" or something else (hyphens and underscores are discouraged from what I understand). This would mean users would do:

```bash
$ pip install uwsift
```

or:

```bash
$ conda install uwsift
```

It would be run by doing the equivalent of:

```bash
$ python -m uwsift ...
```

And inside python, if needed, you would do:

```python
from uwsift.model import Document
```

Note, we could name the **package** "uwsift" but the library is still "sift". This does cause confusion where the user installs "uwsift", but imports and runs "sift".

djhoese:
As #227 says, this PR changes the name for the python package, conda package, and library to `uwsift` instead of `sift`. This is to avoid conflicts with other libraries and packages with a similar name. Similar to @evas-ssec's experience with glance/uwglance.

I think between pycharm and find/replace I got every relevant mention of `sift`.

Closes #227


djhoese:
Workspace clear is required due to Enum imports.

PR #232, opened 2019-08-06
Transition to Satpy for all data reading

djhoese:
This is a start on making SIFT *only* use Satpy for reading input data. There is a lot more left to do though:

- [x] Clean up reader list by showing a human friendly name and the Satpy name. Could read the name from Satpy's configs most likely.
- [x] Show more identifying information in the open file wizard
- [ ] Clean up wizard to be smarter (ex. "Oh these are all single product per file, let's load all those products")
- [ ] Clean up workspace and importers to be more "inviting" to Satpy-like importers. Right now the Scene is created by the wizard and then given to the importer. Maybe the importer could be used directly by the wizard? The importer could provide things like available products (cached or not) and possibly other information. Thoughts @rayg-ssec? This overlaps (in a good way?) with some of the timeline stuff I think.


djhoese:
Just talked with @ScottLindstrom and here is a summary of our discussion:

1. The "properties" tab for choosing multiple datasets should be removed. It causes more confusion that it does benefit users. It was too easy to select levels and names that didn't actually exist in the file (level X was available for name T but not name D; selecting name D and level X would lead to no products).
2. The currently preferred "list" view of products to be loaded should get context menus when you right click on an item's column. If you right click on the name you will see "Select all '<name>'" and "Unselect all '<name>'". Similar menus will be available for each identifying column.
3. There should be a select/unselect all button outside of the list.
4. The final summary page can now be removed since the user *knows* what they are loading since they clicked on the check boxes.

djhoese:
After meeting with @rayg-ssec a few remaining things came up:

1. Fix crash when a new file is added after already viewing the product list (files -> product list -> files (add 1) -> crash).
2. Fix the long_name not being used for reader names. I thought I had included this but it didn't seem to be working for some reason.
3. See if collector can be updated to use the Satpy importer to automatically load metadata for files in a path.

djhoese:
Remaining TODO items will have to wait for a later PR. Too much work going in to one PR at this point.

PR #236, opened 2019-08-19
Use WebEngine instead of building with WebKit

### LINKED ISSUE #233: PyQt5 from conda-forge no longer builds with WebKit
@katherinekolman is back for the semester and had trouble getting SIFT installed in a new environment. One of the issues was that Qt 5.9 on conda-forge is no longer built with WebKit which is what we use to draw the colorbar in the layer details. Qt5 now prefers WebEngine for this type of work. We'll need to fix this usage so that the newest PyQt5 can be used:

https://doc.qt.io/qt-5/qtwebenginewidgets-qtwebkitportingguide.html

katherinekolman:
Closes #233.

This allows the newest version of PyQt5 to be used with SIFT.

### DIFF
--- a/uwsift/view/layer_details.py
+++ b/uwsift/view/layer_details.py
@@ -69,7 +69,7 @@ class SingleLayerInfoPane(QWidget):
         self.clims_text.setSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed)
         self.cmap_vis = QNoScrollWebView()
         self.cmap_vis.setFixedSize(min_width, 30)
-        self.cmap_vis.page().mainFrame().setScrollBarPolicy(Qt.Vertical, Qt.ScrollBarAlwaysOff)
+        self.cmap_vis.page().runJavaScript('document.body.style.overflow = "hidden";')
         self.composite_details = QLabel("Composite Details")
         self.composite_details.setSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed)
         f = QFont()
--- a/conda-recipe/uwsift/meta.yaml
+++ b/conda-recipe/uwsift/meta.yaml
@@ -50,7 +50,7 @@ requirements:
     - pyqtgraph
     - shapely
     - sqlalchemy
-    - pyqt >=5
+    - pyqt >=5.9
     - appdirs
     - pyyaml
     - quamash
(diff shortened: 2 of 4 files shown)


uwsift/view/layer_details.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, document, parent=None):
         self.clims_text.setSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed)
         self.cmap_vis = QNoScrollWebView()
         self.cmap_vis.setFixedSize(min_width, 30)
-        self.cmap_vis.page().mainFrame().setScrollBarPolicy(Qt.Vertical, Qt.ScrollBarAlwaysOff)
+        self.cmap_vis.page().runJavaScript('document.body.style.overflow = "hidden";')
```
djhoese: Is `QNoScrollWebView` used any where else in SIFT? Is it possible to add CSS/styling information to the HTML that gets added to the web view?

katherinekolman: No, it's not used anywhere else in SIFT, but I could add something like:

`<style type="text/css">
body {
    overflow:hidden;
}
</style>`

to the HTML on lines 120 and 125 instead?

djhoese: I guess it might be better the way you have it since this javascript should only run once, right?

djhoese:
Any idea if WebEngine works on PyQt5 5.6? Could you update the version requirement on the `conda-recipes/sift/meta.yaml` to correspond to whatever version is required? I feel like I remember reading that WebEngine was added in 5.3 and WebKit was deprecated in 5.6.

katherinekolman:
I tried testing it with pyqt 5.6.0 and qt 5.6.2, but it doesn't seem to work (can't import PyQt5.QtWebEngineWidgets). It works with 5.9+ though.

uwsift/view/layer_details.py
katherinekolman: Yeah, I think it should only run once.

djhoese:
Sounds good. Qt 5.9 is still pretty old relative to what Qt is available to the public; conda-forge is behind. So, I'm ok with requiring 5.9. This may be an issue when we have to build SIFT if there are any problems with 5.9 that haven't been fixed on all platforms or with pyinstaller (what we use to build the single executable installers).

Nice job figuring this out. Thanks.

djhoese:
Just realized that `pyqt` isn't listed as an `install_requires` in `setup.py`. Could you add this requirement there too?

djhoese:
Looks like the package is called `pyqt5`: https://pypi.org/project/PyQt5/

PR #237, opened 2019-08-20
Fix numba warnings when opening SIFT

### LINKED ISSUE #234: Fix numba warnings when opening SIFT
This is likely a change in numba which shows warnings by default, but @katherinekolman noticed them. Numba is now telling us that our numba.jit functions are not as optimized as they should be. We at least want to get rid of the warnings, but making them optimized would be good too.

katherinekolman:
Closes #234.

I think the numba warnings were because of the [deprecation of object mode fall-back](http://numba.pydata.org/numba-doc/latest/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit), so I just explicitly set `forceobj=True` in a couple places.


djhoese:
Makes sense. This is the easy way to avoid the warning. Can you look in to why it thinks it needs object mode? If changes need to be made to the code, I'm ok with that if it means the code is compiled without object mode.

katherinekolman:
One of the problems was that a 1D array was trying to be indexed with a 2D array, so I changed that. The other problem I'm not sure how to go about is that `np.isclose()` is [not a supported numpy function](https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html) in numba. Is there an alternative function for that?

katherinekolman:
FYI: 2D boolean slicing [is not supported](https://github.com/numba/numba/issues/2538#issuecomment-328605576) by numba but there are ways around it.

uwsift/view/tile_calculator.py
```diff
@@ -0,0 +1,624 @@
+import numpy as np
```
djhoese: Could you add the 2 lines about `#!/usr/bin/env` and encoding? Could you also add a docstring for the module?

PR #238, opened 2019-09-20
Add colorbar option for saving images

### LINKED ISSUE #210: Add option for color bar on exported images
In GitLab by @jgerth on Jan 24, 2019, 12:11

When users export an image or loop from the map display in SIFT, there should be an option to append a color bar horizontally or vertically, with the maximum, minimum, and median value labelled with the corresponding value map. Ideally, this would not overlap with the image. Right now, users must create one from the exported JSON file and append it to the image for publications that require a color bar with figures.

This is an example of one I had to create (ignore that this is a three panel):
gerth_metapps_fig6

### LINKED ISSUE #235: Export image screenshots whole window
Possibly because of PyQt5.

katherinekolman:
Closes #210 and also closes #235. 

This adds a colorbar option when exporting an image. The user can choose to append it horizontally or vertically.

Some screenshots of the colorbar and new export image dialog:

sift_screenshot

sift_screenshot_1

Screen Shot 2019-09-11 at 11 59 09 AM


katherinekolman:
Yeah, all of the colormap files that I added already existed in the python code.

PR #239, opened 2019-09-20
Fix divide by zero error when sizing down GUI

katherinekolman:
This is a minor bugfix that fixes a divide by zero error when sizing the canvas all the way down when an image is loaded.


uwsift/view/visuals.py
```diff
@@ -604,7 +604,7 @@ def get_view_box(self):
         ref_idx_1, ref_idx_2 = get_reference_points(img_cmesh, img_vbox)
         dx, dy = calc_pixel_size(img_cmesh[(self._ref1, self._ref2), :],
                                  img_vbox[(self._ref1, self._ref2), :],
-                                 self.canvas.size)
+                                 self.canvas.size) if self.canvas.size[0] != 0 and self.canvas.size[1] != 0 else (0, 0)
```
djhoese: Maybe this check should be moved to the if statement at the top of the method. Thoughts?

PR #245, opened 2019-10-14
Fix rgb layer creation

katherinekolman:
Prevents crash when trying to make an rgb layer when there are no layers or with another rgb layer, removes numba errors

### DIFF
--- a/uwsift/view/tile_calculator.py
+++ b/uwsift/view/tile_calculator.py
@@ -286,28 +286,23 @@ def visible_tiles(z_dy, z_dx,
     return tilebox
 
 
-@jit([nb_types.UniTuple(nb_types.Tuple([int64, int64, int64]), 2)(
+@jit(nb_types.UniTuple(nb_types.Tuple([int64, int64, int64]), 2)(
     int64,
     int64,
-    nb_types.Tuple([int64, int64]),
-    nb_types.NamedUniTuple(int64, 2, Point),
-    nb_types.NamedUniTuple(int64, 2, Point)),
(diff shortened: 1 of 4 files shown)


uwsift/view/tile_calculator.py
```diff
@@ -376,10 +382,13 @@ def calc_stride(v_dx, v_dy, t_dx, t_dy, overview_stride_y, overview_stride_x):
...
-@jit(nb_types.UniTuple(int64, 2)(
+@jit([nb_types.UniTuple(int64, 2)(
     nb_types.NamedUniTuple(int64, 2, Point),
-    nb_types.NamedUniTuple(int64, 2, Point)
-), nopython=True, cache=True, nogil=True)
+    nb_types.NamedUniTuple(int64, 2, Point)),
+    nb_types.UniTuple(int64, 2)(
+    nb_types.Tuple((int64, int64)),
+    nb_types.NamedUniTuple(int64, 2, Point))
+], nopython=True, cache=True, nogil=True)
```
djhoese: These look like really similar signatures. Could you fix the visual code so they pass the same type of arguments, if possible? I think it would be best if they sent the namedtuples rather than the regular tuples.

katherinekolman: So the functions that were causing the errors were passing in tuples/named tuples, but when I changed the argument's type to match what was in the numba signature originally, it was giving me the same error I was running into before where the argument's type seemed to match the type in the numba signature but numba still threw type errors anyways. I ended up splitting the tuple into its components and it works that way

djhoese: Gross, but ok. Thanks.

PR #246, opened 2019-10-16
Add export image tests

katherinekolman:
This adds tests for export_image.py (not including GUI tests) and a travis configuration file.


setup.py
```diff
@@ -186,9 +186,9 @@ def run(self):
     include_package_data=True,
     install_requires=['numpy', 'pillow', 'scipy', 'numba', 'vispy>=0.6.0',
                       'PyOpenGL', 'netCDF4', 'h5py', 'pyproj',
-                      'pyshp', 'shapely', 'rasterio', 'goesr', 'sqlalchemy',
-                      'goesr', 'appdirs', 'pyyaml', 'pyqtgraph', 'satpy',
-                      'pygrib', 'imageio', 'pyqt5>=5.9'
+                      'pyshp', 'shapely', 'rasterio', 'sqlalchemy',
+                      'appdirs', 'pyyaml', 'pyqtgraph', 'satpy',
+                      'pygrib;sys_platform=="linux" or sys_platform=="darwin"', 'imageio', 'pyqt5>=5.9'
```
djhoese: I don't think I've ever seen this syntax. Is this valid for a `setup.py`? I feel like I've seen it in other requirements listing formats, but not in setup.py like this.

katherinekolman: https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies

is there a better place to put something like this other than in `setup.py`?

djhoese: Looks good. No `setup.py` is the better/best/proper place for this. We use `setup.py` for things that involve pip (ex. `pip install uwsift`) and there is the conda recipe when we build conda packages. I believe the `meta.yaml` for uwsift already has the platform specific dependency for pygrib.

katherinekolman:
pytest finds tests that are named test_*.py or *_test.py but I agree that test_*.py makes more sense, I can change it.

PR #247, opened 2019-10-31
Remove unnecessary band metadata

djhoese:
As SIFT evolves and depends on Satpy more, we no longer have easy access to the band **number** of a specific data variable. Typically what band a variable is is part of the name ("B01", "C01", etc) of that variable. Additionally, the band number is currently used in some metadata (guidebook) operations like "what CF standard_name is this band?" which is unreliable once we support satellites beyond ABI/AHI. Overall, guidebook things should be moved to Satpy as much as possible or should be configurable by the user (once donfig is available).

### DIFF
--- a/uwsift/common.py
+++ b/uwsift/common.py
@@ -275,7 +275,6 @@ class Info(Enum):
 
     # instrument and scene information
     PLATFORM = 'platform'  # full standard name of spacecraft
-    BAND = 'band'  # band number (multispectral instruments)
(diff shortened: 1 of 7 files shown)


rayg-ssec:
I like the idea of relying on satpy to unify this, with a donfig layer to specify what satpy can't/won't, and preferably to handle local overrides (as an "eventually", likely a later ticket?).

PR #248, opened 2019-10-31
Remove quamash as a dependency

djhoese:
Quamash is unmaintained. The next best project that is semi-alive is [asyncqt](https://pypi.org/project/asyncqt/). I am not replacing quamash with asyncqt however because I'm not sure we need it quite yet. We will likely in the future, but I'd like to avoid managing this dependency until we actually need it.

### DIFF
--- a/conda-recipe/uwsift/meta.yaml
+++ b/conda-recipe/uwsift/meta.yaml
@@ -53,7 +53,6 @@ requirements:
     - pyqt >=5.9
     - appdirs
     - pyyaml
-    - quamash
     - satpy >=0.10.0a0
     - pygrib  # [not win]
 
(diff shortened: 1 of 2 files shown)


k30n1:
Agreed on removing an unmaintained module that we're not currently taking advantage of. In the immortal (and thoroughly antiquated) words of the Three's Company opening credits, "Joyce Dewitt"

rayg-ssec:
Agreed that it isn't needed (generators as formal background tasks has worked fine so far) - yet. Easy to revert (or more likely replace) if needed. Proceed

djhoese:
>  Easy to revert

Yeah, that's why I only commented out the quamash stuff instead of deleting it like it never existed. Should be a 1:1 with asyncqt since it is based on quamash.

PR #256, opened 2019-11-15
Add tests for tile_calculator.py

katherinekolman:
This adds tests for the methods in `tile_calculator.py`.

### DIFF
--- a/uwsift/tests/conftest.py
+++ b/uwsift/tests/conftest.py
@@ -1,10 +1,16 @@
 import pytest
 from uwsift.__main__ import Main
 from uwsift.util.default_paths import USER_CONFIG_DIR
+from PyQt5.QtTest import QTest
 
 
 @pytest.fixture(scope="session")
 def window(tmp_path_factory):
     d = tmp_path_factory.mktemp("tmp")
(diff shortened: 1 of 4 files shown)


djhoese:
Nice work. I love a good coverage increase.

Looks like the coverage can't cover the numba jit'd stuff. A quick googling makes it seem like we can either disable the jit'd-ness during coverage tests or just accept that the numba jit functions aren't included. I don't really like the idea of turning the jit off for coverage unless we run the tests twice (once with jit, once without) because otherwise we aren't actually testing their validity.

What do you think @katherinekolman?

katherinekolman:
Running the tile_calculator tests twice with jit on and off doesn't add much overhead to test runtime (a few seconds at most) and only adds an extra percent to the overall coverage score. I'd have to run the tile_calculator test separately though again so the travis.yml would look something like this:

```
- pytest --cov=uwsift tests/
- NUMBA_DISABLE_JIT=1 pytest --cov=uwsift tests/.../test_tile_calculator.py
```

If you're fine with this I can add it.

djhoese:
What about a pytest decorate to parameterize the test (I think that's what they call it) with a boolean value for whether or not numba is mocked?

PR #262, opened 2019-11-25
Remove timeline from user interface until feature is complete

djhoese:
This removes the timeline view from the main window user interface. The feature was not complete, took up space, and mostly confused users since it wasn't really documented. @rayg-ssec has said in the past that he would like to rewrite a lot of it so I think it is best to remove it from the user's view.

The nice thing about the way I did it in this PR is that I didn't have to change the `.ui` files.

### DIFF
--- a/uwsift/__main__.py
+++ b/uwsift/__main__.py
@@ -746,8 +746,8 @@ class Main(QtGui.QMainWindow):
         self._init_metadata_background_collection(search_paths)
 
         # set up timeline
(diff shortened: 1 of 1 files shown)


rayg-ssec:
Looks like a clean and simple disable. Agree with removing this UI element - in addition to already mentioned rationale, timeline UI needs to be rewritten for Qt5 (e.g. QML, possibly repurposed from a pre-existing open source nonlinear editor application). QGraphicsScene and its use of OpenGL is headed into deprecation.

PR #263, opened 2019-11-25
Add caching to satpy available readers

djhoese:
This adds an on-disk file to cache what readers Satpy has available to the user. It does this by loading the reader information from Satpy, adding it to a dictionary along with the current `satpy_version`. If the file doesn't exist, the contents are empty, or the satpy version is older than the currently imported satpy then the reader information is regenerated and re-cached.

This cuts off a noticeable 2-4 seconds from loading SIFT.

## TODO

- [x] Add tests

### DIFF
--- a/uwsift/tests/workspace/test_importer.py
+++ b/uwsift/tests/workspace/test_importer.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Tests for the importer functions and classes."""
+
+import os
+import yaml
+from uwsift.workspace.importer import available_satpy_readers
+
+
+def test_available_satpy_readers_defaults():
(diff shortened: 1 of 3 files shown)


rayg-ssec:
Reviewed code changes and am fine with the merge. Only question that comes to mind is whether the cache clear flag for startup will also be funneled into the cache clear option on this new code. Given that the file location is being logged, and the primary parties that may want to force a reset are devs, I think it strictly optional/undefined whether the reader list is reset with the workspace cache.

djhoese:
A command line flag would be nice, but I didn't want to drag it through all of the components that call (or may call in the future) this function. With the donfig PR merged I could do this as a global configuration option. I still see the workspace and this file as separate things that are being cached and would like to not make the one flag work for both items.

PR #270, opened 2019-12-05
Fix open file wizard when resolution is a floating point number

### LINKED ISSUE #265: `_pretty_identifiers` crash if `resolution` is a float
Some Satpy-readers have instrument `resolution` defined as a float. String formatting for:
`pretty_val = "{:d}m".format(value)` crashes as it is expecting an `int`.

djhoese:
Closes #265 

Some how I got lucky and assumed that `resolution` from Satpy was always an integer. This forces resolutions to an integer number to be formatted in the open file wizard dialog.

### DIFF
--- a/uwsift/__main__.py
+++ b/uwsift/__main__.py
@@ -20,7 +20,6 @@ REQUIRES
 
 __author__ = 'rayg'
 
-import asyncio
 import logging
 import os
 import sys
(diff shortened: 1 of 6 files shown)


djhoese:
I made an issue about the odd pyqt enum behavior: https://github.com/conda-forge/pyqt-feedstock/issues/72
It seems to only be a problem with the conda-forge build, but only effects our tests so my current workaround is good enough for us.

Merging...

PR #281, opened 2020-02-13
Add resampling functionality to SIFT

katherinekolman:
This adds a menu that allows the user to resample data.

Screen Shot 2020-02-13 at 2 54 06 PM

Some VIIRS data loaded in.

Screen Shot 2020-02-13 at 2 46 45 PM

The resample button is located in the lower right corner.

Screen Shot 2020-02-13 at 2 47 03 PM

When you click on the Resample Data button, this menu appears.

Some notes:
- Only swath data is able to be resampled. If data with area definitions are loaded and the user tries to resample the data, SIFT will ignore it.
- If swath data is being loaded in and the user doesn't specify any resampling options, it will default to using nearest neighbor and the projection currently shown on the screen.
- The new area definition that the data is resampled to is made from `scene.max_area()`


djhoese:
I'd like to merge this, but having this resampling UI stuff added without explanation or proper training time might be an issue. Do you think you could add an option to the config as to whether or not this feature is enabled and if disabled disable the button in the UI?

djhoese:
Could you move the resample config option under the open_file_wizard?

djhoese:
This was merged/implemented as part of the work by EUMETSAT and recently merged in the all-in-one PR #341. Thanks again Kat.

PR #282, opened 2020-03-03
Add SIFT script to python package installation

djhoese:
This PR adds an `entry_point` to the `setup.py` file so when the `uwsift` package is installed in an environment a `bin/SIFT` script will be made available to start SIFT.

This is the first step in producing a conda-pack tarball for linux-installations. The current instructions (not in the official docs yet) are:

```
conda-pack --arcroot SIFT_1.1.2 --output /tmp/SIFT_1.1.2.tar.gz
```

with a valid conda environment already activated. Note `uwsift` must be installed with `pip install --no-deps` instead of the more typical development environment install of `pip install -e .`. The important part is the `-e`. Doing it this way actually moves the uwsift files to the python environment instead of just linking them.

Once we have this tarball we can then do the following to install it on another system:

```
tar -xzf SIFT_1.1.2.tar.gz
source SIFT_1.1.2/bin/activate
SIFT ...
```

CC @katherinekolman Could you test this branch and build instructions on your Mac?


djhoese:
We've waited long enough, let's merge this. @katherinekolman can continue work on understanding the matplotlib test failure in #286.

PR #283, opened 2020-03-05
Remove test case for colorbar creation

katherinekolman:
This removes a test that tests a canvas of size [0, 0] creates a colorbar of size [0, 0] since it wasn't working with matplotlib 3.2.0.


djhoese:
Failures are from the Windows PyQt issue that we still haven't figured out. Looks good.

PR #284, opened 2020-03-11
Test DLL failures on Travis CI Windows environment

djhoese:
This is an evolution of the problems popping up on Travis and that I've been testing in #282. I'm making this so I can continue testing things while I accept defeat on the Windows builds going forward in normal SIFT development.


djhoese:
I'm not sure how I forgot to close this, but this has been figured out although with an undesired result. Travis Windows does not support OpenGL/virtual displays. This is mentioned here:

https://travis-ci.community/t/make-running-gui-multimedia-apps-available-in-windows/1557

and can be voted on to try to move it higher in their priority list. So, for now, the Windows tests can't be run on Travis CI and have been skipped as part of #282.

PR #287, opened 2020-05-20
[WIP] First steps in fixing SIFT to auto-flip SEVIRI data

djhoese:
This is mainly here so I don't lose this and when someone brings it up and I say "I'm pretty sure I already tried to make this work" I can easily find the code.

This PR attempts to change the handling of geolocation information to allow data that is flipped along one or more axes. This is the case for SEVIRI data in most formats (x and y dimensions are flipped) and FCI (y dimension is flipped). However, this currently doesn't work perfectly. I originally thought this wasn't going to be that hard to I gave it a shot, but ran in to a lot of complications when it came to figuring out where an image's tiles are (ex. where is tile (0, 0)). A lot of the changes were pretty straight forward; stop doing absolute value of the pixel resolution everywhere and change all `-` in the `y` dimension to `+` to match what is being done for the `x` dimension. This mostly worked, but the navigation is still off when displayed on the map:

Screenshot_20200508_163459

### DIFF
--- a/uwsift/view/tile_calculator.py
+++ b/uwsift/view/tile_calculator.py
@@ -111,9 +111,9 @@ def calc_pixel_size(canvas_point, image_point, canvas_size):
     # (1 - (-1) = 2). Use this ratio to calculate number of
     # screen pixels between the two reference points. Then
     # determine how many image units cover that number of pixels.
-    dx = abs(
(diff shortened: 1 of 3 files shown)


djhoese:
CC @sjoro

PR #293, opened 2020-05-21
Fix HiDPI setting being set too late

djhoese:
The HiDPI setting for Qt5 needs to be set before the QApplication is created. This PR does that along while also rearranging when the app is created.

### DIFF
--- a/uwsift/__main__.py
+++ b/uwsift/__main__.py
@@ -55,11 +55,6 @@ from uwsift.view.scene_graph import SceneGraphManager
 from uwsift.workspace import Workspace
 from uwsift.workspace.collector import ResourceSearchPathCollector
 
-app_object = app.use_app('pyqt5')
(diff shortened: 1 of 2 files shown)


rayg-ssec:
Note that this attribute is a Windows/Linux setting and that macOS is likely to need a different solution. That should probably be done as a separate merge.

djhoese:
Do you get the warning even with this branch @rayg-ssec? If this is a no-op on macOS then we could still merge this right? But then later figure out the other solution (.plist or other).

djhoese:
Interestingly...this cause a segfault on my local machine and on travis. Not sure why.

rayg-ssec:
ca3c6812115a0537ada4ed2fd91dec165fbab43f on macOS no longer generates warnings on startup.

rayg-ssec:
Likewise with d29ea9f .

Issue #295, opened 2020-05-27
Update bundle scripts to handle directory being moved by user


djhoese:
Right now the two bundle scripts (SIFT.bat and SIFT.sh) only check if the script has been run before by checking a `.installed` file. While this stops `conda-unpack` from being run multiple times it doesn't fix the case where a user runs SIFT once and then moves the directory. In this case `conda-unpack` should probably be run again to re-write all the directories (assuming conda-unpack can rewrite paths multiple times).

We should update both of these scripts to check the path that is in the file, see if it matches the base directory we are currently in and if not then rerun conda-unpack and rewrite the file.

@katherinekolman I thought you could do the `.bat` script and I'll do the `.sh` script.

PR #296, opened 2020-05-27
Update bundle scripts to handle installation directory being moved

djhoese:
Attempt at fixing #295. At the time of writing this is just the `.sh` script, but should work.

I started researching `.bat` file solutions and came across this: https://stackoverflow.com/questions/130116/windows-batch-commands-to-read-first-line-from-text-file. I'm not sure how Windows handles newlines and carriage returns the way @katherinekolman has it writing to the file. Kat do you think you could look in to this and test it on Windows? The idea is that if you run SIFT once, then move the directory to a new directory and run it again, conda-unpack should run and data should load fine.

### DIFF
--- a/bundle_scripts/SIFT.bat
+++ b/bundle_scripts/SIFT.bat
@@ -8,11 +8,19 @@ call %base_dir%Scripts\activate
 
 REM Create a signal file that we have run conda-unpack
 set installed=%base_dir%.installed
-if not exist "%installed%" (
+if not exist "%installed%" goto install
+
+set /p install_dir=< %installed%
+if not %base_dir% == %install_dir:~0,-1% goto install
+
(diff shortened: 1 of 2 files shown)


katherinekolman:
I updated the .bat script and tested it on Windows, but I'm not sure why the tests are failing like this...

djhoese:
Looks like a new version of pyvirtualdisplay is breaking things for pytest xvfb: https://github.com/The-Compiler/pytest-xvfb/issues/22

Let's give it 24 hours and then see where things stand. We can limit the version of pyvirtualdisplay to <1.0 if we have to.

djhoese:
A new version of pytest-xvfb just came out, I restarted the tests, and now things seem to be passing.

Issue #297, opened 2020-06-12
SIFT 1.1.3 windows experimental builds do not start


rayg-ssec:
Downloaded 20200612_160153 experimental version, and unpacked it onto a Win10pro VM with OpenGL3 drivers to confirm that it starts up. Noted that the EXEs and DLLs are all placed in the SIFT_1.1.3 directory instead of the DLLs or bin directory, which may or may not be significant to the below error.

Running SIFT startup script appeared to initially work as intended, then failed. Intercepting the error before the window closed was somewhat difficult. A partial screengrab is included showing "The application failed to start because no Qt platform plugin could be initialized." after 'Could not find the Qt platform plugin "windows" in ""'.

<img width="992" alt="image" src="https://user-images.githubusercontent.com/1828555/84553284-62947a80-acd9-11ea-8e5b-1c7651020a9f.png">

rayg-ssec:
Further illustration of the zipfile layout. Note that I used windows explorer to unpack the zipfile, all 85000+ files of it. 

<img width="868" alt="image" src="https://user-images.githubusercontent.com/1828555/84553431-e9e1ee00-acd9-11ea-8f32-51b9da827c92.png">

<img width="688" alt="image" src="https://user-images.githubusercontent.com/1828555/84553476-fc5c2780-acd9-11ea-8af3-23f605e5e7b4.png">

djhoese:
One thing to try: maybe the environment isn't being activated properly in the `.bat` file. You could try manually activating it then calling `python -m uwsift`. However, I'm not sure the proper way to activate a conda-pack'd environment on Windows because it isn't documented in their docs from what I can see:

https://conda.github.io/conda-pack/

rayg-ssec:
The issue is that the qt.conf in the base directory needs updating. 

as shipped:
```
[Paths] 
Prefix = C:/tools/miniconda3/envs/test/Library
Binaries = C:/tools/miniconda3/envs/test/Library/bin
Libraries = C:/tools/miniconda3/envs/test/Library/lib
Headers = C:/tools/miniconda3/envs/test/Library/include/qt
TargetSpec = win32-msvc
HostSpec = win32-msvc
```

Example of a patched version which allowed SIFT to start up:
```
[Paths] 
Prefix = C:/Users/user/Desktop/SIFT_1.1.3dev_windows_20200612_160153/SIFT_1.1.3/Library
[...]

PR #300, opened 2020-07-29
297 Relocate qt.conf after conda-unpack for Windows

rayg-ssec:
This fixes inability to start the bundled form of SIFT on Windows. Tested in a VM by manually adding to SIFT 1.1.3 test bundle, so it may yet need addition to any manifest for the bundle build (have not found one so far).
- SIFT.bat now uses %~dp0 to include drive letter of the install location.
- qt_conf_fix.py is run by SIFT.bat at the time that conda-unpack is completed (first run).
- qt_conf_fix.py replaces Qt Library locations in qt.conf, required in order to start up.
- qt_conf_fix.py could be used by Unix OSes if also needed there, it takes the install directory base as its singular argument.

### DIFF
--- a/bundle_scripts/SIFT.bat
+++ b/bundle_scripts/SIFT.bat
@@ -1,7 +1,7 @@
 @echo off
 REM Initialize SIFT installation if necessary and run SIFT
 
-set base_dir=%~p0
+set base_dir=%~dp0
 
 REM Activate the conda environment
 call %base_dir%Scripts\activate
--- a/bundle_scripts/SIFT.bat
+++ b/bundle_scripts/SIFT.bat
@@ -11,6 +11,7 @@ set installed=%base_dir%.installed
 if not exist "%installed%" (
   echo Running one-time initialization of SIFT installation...
   conda-unpack
+  python %base_dir%\qt_conf_fix.py %base_dir%
   echo %base_dir% > %installed%
 )
 
--- a/bundle_scripts/qt_conf_fix.py
+++ b/bundle_scripts/qt_conf_fix.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+# Relocate qt.conf after conda-unpack for Windows (see SIFT.bat)
+# example qt.conf file before relocation:
+#
+# [Paths]
+# Prefix = C:/tools/miniconda3/envs/test/Library
+# Binaries = C:/tools/miniconda3/envs/test/Library/bin
+# Libraries = C:/tools/miniconda3/envs/test/Library/lib
+# Headers = C:/tools/miniconda3/envs/test/Library/include/qt
+# TargetSpec = win32-msvc
+# HostSpec = win32-msvc
+
+import sys
+import os
+import re
+
+
+def main(base):
+    fn_qt_conf = os.path.join(base, 'qt.conf')
+    with open(fn_qt_conf, 'rt') as fob:
+        txt = fob.read()
+        fob.close()
+    src, = re.findall(r'Prefix\s*=\s*(.*?)Library', txt, re.MULTILINE)
+    dst = base.replace('\\', '/')
+    if not dst.endswith('/'):
+        dst += '/'
+    print(f"Relocating qt.conf from {src} to {dst}")
+    new_txt = txt.replace(src, dst)
+    with open(fn_qt_conf, 'wt') as fob:
+        fob.write(new_txt)
+        fob.close()
+
+
+if __name__ == '__main__':
+    main(*sys.argv[1:])
+
--- a/bundle_scripts/qt_conf_fix.py
+++ b/bundle_scripts/qt_conf_fix.py
@@ -33,4 +33,3 @@ def main(base):
 
 if __name__ == '__main__':
     main(*sys.argv[1:])
-


djhoese:
Nice, thanks. Did you ever check if linux has this conf file? I wonder if this is a bug in conda-pack that should be fixed where it isn't detecting the paths in the qt.conf file on Windows.

djhoese:
So I just downloaded the 1.1.3 tarball, unpacked it, and looked at `bin/qt.conf`. `conda-unpack` is actually fixing things on the linux side:

```
(base) davidh@janet:/tmp/SIFT_1.1.3$ cat bin/qt.conf
[Paths]
Prefix = /opt/anaconda1anaconda2anaconda3
Binaries = /opt/anaconda1anaconda2anaconda3/bin
Libraries = /opt/anaconda1anaconda2anaconda3/lib
Headers = /opt/anaconda1anaconda2anaconda3/include/qt

(base) davidh@janet:/tmp/SIFT_1.1.3$ bin/conda-unpack

(base) davidh@janet:/tmp/SIFT_1.1.3$ cat bin/qt.conf
[Paths]
Prefix = /tmp/SIFT_1.1.3
Binaries = /tmp/SIFT_1.1.3/bin
Libraries = /tmp/SIFT_1.1.3/lib
Headers = /tmp/SIFT_1.1.3/include/qt
```

djhoese:
Side question: Where did `297` come from in the title?

djhoese:
Ok best I can see this is a problem because on Windows `qt.conf` is in the root directory instead of the `bin` directory. Here is the conda-unpack portion of `conda-pack`:

https://github.com/conda/conda-pack/blob/ea74ae159ec0f1937979ac7fa0cff112c59e77d4/conda_pack/core.py#L949-L951

```
        if ((self.has_dest or
             file_mode == 'unknown' or
             file_mode == 'text' and file.target.startswith(BIN_DIR))):
```

So it probably detects a text file that isn't in the bin dir and skips over it.

djhoese:
So found some new evidence that this may be an unfortunate configuration choice of conda-forge. They seem to create the qt.conf file in `<env_base>/Library/bin/` when you activate the environment:

https://github.com/conda-forge/qt-feedstock/blob/master/recipe/write_qtconf.bat

Here is the qt.conf they use for linux but not sure this is the same one that is used on Windows:

https://github.com/conda-forge/qt-feedstock/blob/master/recipe/qt.conf

rayg-ssec:
297 comes from the ticket number. 

Do we know if there's a deliberate reason these files are getting splashed into the base directory when unpacking, e.g. softlinks that are being added, which are getting turned into copies during archival? And if that's the case, is the better solution to ensure that the right binaries, scripts, config files, and DLLs are being used (from Library/bin rather than base)?

djhoese:
Closes #297.

There now it is linked.

I have a feeling it is softlinks or something of the sort. In what I'm seeing `qt.conf` is the most out of place thing in the base directory. I think I'll create an issue with the conda-forge qt people and ask about the proper locations for these things *or* talk to conda-pack about modifying (or allowing additional) BIN search directories. Let me double check conda-pack first.

djhoese:
It would seem this PR, with the custom solution for updating qt.conf, is not needed and instead we could run:

```
<env>\Scripts\.qt-post-link.bat
```

which is the conda-forge activation script for the Qt library. Why it isn't run in the conda-pack environment I'm not sure yet.

djhoese:
They are apparently supposed to be run as part of conda's activation: https://docs.conda.io/projects/conda-build/en/latest/resources/link-scripts.html

djhoese:
Ok I have finally figured this out. I couldn't let this problem go. This is not necessary a bug in conda-forge's decisions or in conda-pack but rather in a limitation of conda-pack for handling certain paths on Windows. The paths in qt.conf as-is are:

```
[Paths] 
Prefix = C:/Users/User/Miniconda3/envs/conda_unpack_test/Library
Binaries = C:/Users/User/Miniconda3/envs/conda_unpack_test/Library/bin
Libraries = C:/Users/User/Miniconda3/envs/conda_unpack_test/Library/lib
Headers = C:/Users/User/Miniconda3/envs/conda_unpack_test/Library/include/qt
TargetSpec = win32-msvc
HostSpec = win32-msvc
```

conda-pack searches for an old prefix with `\\` (a single backslash) but replaces it with a single forward slash. As you can see the existing file produced by the post-link script already has forward slashes. So the find/replace that conda-unpack does never replaces these paths. One solution is to edit or add these lines:

```
('qt.conf', 'C:\\Users\\User\\Miniconda3\\envs\\conda_unpack_test', 'text'),
('Library\\bin\\qt.conf', 'C:\\Users\\User\\Miniconda3\\envs\\conda_unpack_test', 'text')
```

with these lines:


```
('qt.conf', 'C:/Users/User/Miniconda3/envs/conda_unpack_test', 'text'),
('Library\\bin\\qt.conf', 'C:\\Users\\User\\Miniconda3\\envs\\conda_unpack_test', 'text')
```

In `<unpacked_env>/Scripts/conda-unpack-script.py`. Doing this allows me to run this without error:

```
python -c "from PyQt5 import QtGui, QtWidgets; app = QtWidgets.QApplication([]); w = QtWidgets.QWidget(); w.show(); app.exec_()"
```

djhoese:
Or...since we have control of the environment before conda-packing it, we could find/replace these exact paths in qt.conf to use `\\` before running conda-pack. That way conda-unpack would find the prefix it is looking for. I will file a bug with conda-pack to ask for other ideas.

djhoese:
https://github.com/conda/conda-pack/issues/141

djhoese:
Supersceded by #302

PR #301, opened 2020-08-03
Add support for Satpy 0.23+ and PROJ 6.0+

djhoese:
This PR fixes two huge updates from two of our dependencies:

1. In Satpy PR https://github.com/pytroll/satpy/pull/1088, @mraspaud has migrated from a single over-purposed `DatasetID` object to `DataID` objects and `DataQuery` objects. Conceptually these should help separate the use cases of "I know exactly what I have and here is how it is identified" and "I don't know exactly what I want, but I'd like to query for these parameters". Functionally this requires a couple changes, but the main one is that DataID properties should be access as dictionary keys, not properties. This PR tries to support both old property/attribute access and dictionary getitem access to these properties. I've done some basic testing and things seem to work.
2. Pyproj 2.x and PROJ 6.0+ no longer support the `+over` parameter. This was a "hack" that allowed us in the SIFT world to project coordinates that were outside of -180/180 space and pretend like they were in that space. This parameter is still used in SIFT's transformation code, but other areas needed to be updated. Mainly, handling of grib data that is 0-360 longitude is now swapped around to be -180/180 to fit in the primary SIFT viewing domain. I believe this changes how grib files were viewed previously which repeated the data from -180 to ~540 degrees longitude. This is no longer possible without some larger modifications to how SIFT handles projections and data (ex. making two copies of each image so it appears on both -180/180 and 180/540). Note that for all of this to work depends on yet another Satpy PR to fix the grib reader (I'll make this shortly).

TODO:

- [x] Add tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ script:
 - python build_conda_pack.py -j -1 $oflag
 - ls -l
 after_success:
-- if [ $PYTHON_VERSION == "3.7" ] && [ "${TRAVIS_OS_NAME}" != "windows" ]; then coveralls;
+- if [ $PYTHON_VERSION == "3.7" ] && [ "${TRAVIS_OS_NAME}" != "windows" ]; then coveralls; fi
 - if [[ $TRAVIS_TAG == "" ]]; then
     odir="experimental/";
   else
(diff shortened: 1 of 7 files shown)


djhoese:
Related Satpy PR: https://github.com/pytroll/satpy/pull/1296

PR #302, opened 2020-08-06
Add workaround to qt.conf in conda-pack

### LINKED ISSUE #297: SIFT 1.1.3 windows experimental builds do not start

djhoese:
Closes #297. Replaces #300.

### DIFF
--- a/build_conda_pack.py
+++ b/build_conda_pack.py
@@ -66,7 +66,7 @@ def main():
         raise RuntimeError(f"Unknown platform: {sys.platform}")
 
     # HACK: https://github.com/conda/conda-pack/issues/141
-    if 'win' in sys.platform:
+    if sys.platform.startswith('win'):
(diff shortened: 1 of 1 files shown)


rayg-ssec:
Agree that fixing conda-pack logic is a better option than SIFT.bat - advise continuing with this and dropping https://github.com/ssec/sift/pull/300

Issue #308, opened 2021-01-11
win32 packaging of sift 1.1.4 does not auto-update qt.conf


rayg-ssec:
After doing some diagnostics, @djhoese and I conclude that the qt.conf resulting from the build process is not properly triggering the generation of accurate qt.conf during conda-unpack phase.

Fix `build_conda_pack.py` such that
- Prefix entry in qt.conf in the resulting tar ball uses \
- all other entries in the tar ball's qt.conf have identical prefix with \, but use / for subsequent path components

During unpack, prefix gets replaced throughout all entries with a correct installation path, but using / instead of \. However, components after the prefix are repeated verbatim.

ameraner:
Closing as solved

PR #310, opened 2021-01-11
Replace / with \\ throughout Prefix where present in qt.conf, to fix …

rayg-ssec:
Partially tested qt.conf update during packaging to address #308 
- Extract Prefix string
- replace / with \\ in Prefix
- replace prefix string with new version throughout qt.conf configuration file

Also now uses text-mode open() when reading and writing qt.conf.

This is intended to allow conda-unpack to properly detect and amend Prefix on Windows.

### DIFF
--- a/build_conda_pack.py
+++ b/build_conda_pack.py
@@ -19,6 +19,7 @@ import os
 import sys
 import shutil
 import subprocess
+import re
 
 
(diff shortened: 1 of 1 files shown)


djhoese (approved):
Looks good to me. Thanks!

PR #311, opened 2021-01-11
Retain forward slash before Libary in prefix

rayg-ssec:
Revised for #308 after 1.1.5 found unviable for win32


djhoese:
Third times a charm.

PR #312, opened 2021-06-22
Update to work with VisPy 0.7

djhoese:
VisPy 0.7 is on its way to being released (hopefully this week) and includes some major changes to the ImageVisual which SIFT uses heavily (also big changes to VolumeVisual and MeshVisual but that doesn't matter as much to SIFT). Back in the day, I had to do some heavy hacking to get SIFT to be able to use VisPy's ImageVisual the way I wanted (gamma, floating-point textures, etc). These hacks are not compatible with the new version of the ImageVisual. The good news is that the new version of the ImageVisual includes all of these OpenGL-specific features that I was using (but not the geolocation-specific stuff). This PR updates the SIFT version of the ImageVisual to work with VisPy 0.7. Because of the changes needed, VisPy 0.7 will be the new minimum version of vispy for SIFT (no older versions).

TODO:

- [x] Check if the NaN check is needed in the fragment shader. If not, we can probably get rid of our custom shaders which will be awesome.
- [x] Update the composite/RGB Visual to work with VisPy 0.7. If possible separate functionality into Mixin classes (one for geolocation, one for multi-band ImageVisual). This will simplify making a multi-band ImageVisual for VisPy in the future.
- [x] Test. Test. Test.


djhoese:
So I think this PR is ready for merge. I wanted to include all this work in one PR so that it was easier for @sjoro and his contractors to copy it over to their instance of SIFT. This PR involves major restructuring of the Image-based Visuals in SIFT and needs *a lot* of testing. With the basic tests that I've done it seems to be working.

This PR separates things out as much as possible including creating a new MultiChannelImageVisual class (and associated texture wrapper class and tests) with my plan being to include it in vispy upstream in something like version 0.8 or 0.9 (assuming the other maintainers are OK with it).

@rayg-ssec what is your opinion on merging this and start work on GitHub actions so that we can get some beta tarballs to @ScottLindstrom and maybe @jgerth if he has the time? Or would you like to test this yourself before merging?

djhoese:
Note to @sjoro, this PR now also includes changes by @rayg-ssec to fix compatibility with pyqtgraph 0.12.2+ where some Qt imports were getting mixed together and letting us do some bad things.

PR #316, opened 2021-07-23
Improve log message when tiled image isn't displayed

djhoese:
Original commit message:

```
When the TiledGeolocatedImage is panned to leave the canvas completely a
ValueError exception is raised to notify about this and also a WARNING log
message is generated (it can be discussed, whether this case is worth a
warning, but...). This commit enhances the wording and adds the
ValueError exception message to the log message in the hope to reduce
confusion about its meaning.
```

This is the first commit taken directly from the EUMETSAT gitlab that was contributed by their contractors. It makes sense and is a simple fix. This is a cherry-pick so it completely breaks commit history with their version, but that's going to have to happen if we only merge specific pieces (I'd like to merge all of it).


djhoese:
Original gitlab commit: https://gitlab.eumetsat.int/webservices/mtg-sift/-/commit/c8055b56879d5b0636a6c4dd2e00f14015a424be

PR #321, opened 2021-08-03
Add initial basic github actions

djhoese:


djhoese:
Did a test as part of this PR and looks like tarball upload to SSEC FTP works. The only untested part of this CI is:

1. PyPI package upload. I can't test this really until we make a release. I'm not too worried about this since I copied it from the Satpy CI configs.
2. I don't test on OSX or Windows. It would take more work to get OpenGL to work. Since VisPy currently doesn't test it I decided not to try it here.

I will merge once the current tests pass.

djhoese:
Also note that one of the environments has experimental/unstable versions of xarray and dask. This should hopefully make it faster to detect incompatibility issues.

djhoese:
Ok I made one more update. I made it so Satpy is included in the unstable dependencies.

djhoese:
Crap I forgot to delete the renamed travis config.

PR #324, opened 2021-08-04
Update color limit sliders to update display instantly

### LINKED ISSUE #323: Color limit sliders only respond to direct click events
If you change the color limits of either a layer or an RGB component layer, the slider only responds to clicking and dragging the "button" on the slider. Scrolling or clicking on a different part of the slider only move its without actually effecting the display. The animation slider seems to handle this behavior fine.

djhoese:
Closes #323 

Turns out I implemented the previous way on purpose to avoid over taxing the CPU/GPU to draw the updates. However, looks like it handles this just fine. This PR updates the color limit sliders for the individual layers and for the RGB layers so that scrolling or clicking somewhere on the slider range will automatically the visual. Previously this required clicking and releasing the handle on the slider.

### DIFF
--- a/uwsift/tests/view/test_colormap_dialogs.py
+++ b/uwsift/tests/view/test_colormap_dialogs.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Test changes through the colormap dialog."""
+from uwsift.common import Info, Kind, Presentation
+from uwsift.model.layer import DocBasicLayer
+from uwsift.view.colormap_dialogs import ChangeColormapDialog
+from unittest import mock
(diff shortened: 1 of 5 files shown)


rayg-ssec:
Reviewed code changes and spun it up on macOS. Very minor: scene_graph:change_layers_color_limits:L908 should be a debug (or maybe one step beyond) and not an info message after this change, since it gets pretty chatty during slider dragging.

djhoese:
Should I also change the gamma and colormap to debug? They don't change as fast but they are the same level of change.

djhoese:
+1.9% coverage. Woot woot.

All issues have been addressed. Merging now.

PR #331, opened 2022-01-16
Fix Python 3.10 incompatibilities

rayg-ssec:
A number of python 3.10 gotchas including library location changes, type markup, Qt call signature types, float vs integer division prevent SIFT from starting up when using python 3.10.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -54,6 +54,23 @@ autodoc_mock_imports = [
 seqdiag_fontpath = os.path.abspath('./source/fonts/DejaVuSerif.ttf')
 
 
+def _skip_pyqtsignal_doc(app, what, name, obj, skip, options):
+    # see https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-skip-member
+    if what == "class" and "pyqtSignal" in obj.__class__.__name__:
+        return True
+    # fallback to other handlers or autodoc default
+    return None
+
+
+def setup(app):
+    """Global setup function for sphinx documentation.
+
+    Connect to autodoc event so invalid pyqtSignal docs can be skipped.
+
+    """
+    app.connect('autodoc-skip-member', _skip_pyqtsignal_doc)
+
+
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
 
@@ -228,25 +245,25 @@ htmlhelp_basename = 'SIFTdoc'
 # -- Options for LaTeX output ---------------------------------------------
 
 latex_elements = {
-# The paper size ('letterpaper' or 'a4paper').
-#'papersize': 'letterpaper',
+    # The paper size ('letterpaper' or 'a4paper').
+    #'papersize': 'letterpaper',
 
-# The font size ('10pt', '11pt' or '12pt').
-#'pointsize': '10pt',
+    # The font size ('10pt', '11pt' or '12pt').
+    #'pointsize': '10pt',
 
-# Additional stuff for the LaTeX preamble.
-#'preamble': '',
+    # Additional stuff for the LaTeX preamble.
+    #'preamble': '',
 
-# Latex figure (float) alignment
-#'figure_align': 'htbp',
+    # Latex figure (float) alignment
+    #'figure_align': 'htbp',
 }
 
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-  (master_doc, 'SIFT.tex', 'SIFT Documentation',
-   'SIFT Developers', 'manual'),
+    (master_doc, 'SIFT.tex', 'SIFT Documentation',
+     'SIFT Developers', 'manual'),
 ]
 
 # The name of an image file (relative to this directory) to place at the top of
@@ -289,9 +306,9 @@ man_pages = [
 # (source start file, target name, title, author,
 #  dir menu entry, description, category)
 texinfo_documents = [
-  (master_doc, 'SIFT', 'SIFT Documentation',
-   author, 'SIFT', 'One line description of project.',
-   'Miscellaneous'),
+    (master_doc, 'SIFT', 'SIFT Documentation',
+     author, 'SIFT', 'One line description of project.',
+     'Miscellaneous'),
 ]
 
 # Documents to append as an appendix to all manuals.
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,7 +43,7 @@ jobs:
         with:
           miniforge-variant: Mambaforge
           miniforge-version: latest
-          python-version: 3.9
+          python-version: 3.10
           use-mamba: true
           environment-file: continuous_integration/environment.yaml
           activate-environment: test-environment
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,6 +43,7 @@ jobs:
         with:
           miniforge-variant: Mambaforge
           miniforge-version: latest
+          python-version: 3.9
           use-mamba: true
           environment-file: continuous_integration/environment.yaml
           activate-environment: test-environment
@@ -109,6 +110,7 @@ jobs:
           miniforge-variant: Mambaforge
           miniforge-version: latest
           use-mamba: true
+          python-version: ${{ matrix.python-version }}
           environment-file: continuous_integration/environment.yaml
           activate-environment: test-environment
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,7 +43,7 @@ jobs:
         with:
           miniforge-variant: Mambaforge
           miniforge-version: latest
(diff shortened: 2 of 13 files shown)


djhoese:
Looks like the website build errors are not caused by your changes:

>  docstring of uwsift.queue.TaskQueue.didMakeProgress:13:Inline emphasis start-string without end-string.

The regular test failures do seem related at least a little bit. Looks like pyqt was updated on conda-forge 2 months ago. Some/most of these errors could be a change in PyQt API which seems dumb (like `setAlignment` not taking the `Alignment` enum).

I don't see any issues on the pyqt feedstock about these issues so my assumption is that they are expected.

djhoese:
For the record, here's what changed when I updated my environment (previous environment completely passed):

```
  sphinx                                         3.0.3-py_0 --> 4.4.0-pyh6c4a22f_1
  sphinxcontrib-htm~                             1.0.3-py_0 --> 2.0.0-pyhd8ed1ab_0
  sphinxcontrib-ser~                             1.1.4-py_0 --> 1.1.5-pyhd8ed1ab_1
```

But it didn't update pyqt/qt. this is a Python 3.7 environment so it must not be getting updates any more from conda-forge.

djhoese:
And Python 3.9 environment when updating a few deps:

```
  pyproj                               3.1.0-py39ha9a7ae0_3 --> 3.3.0-py39hab5ddba_1
  pyqt                                5.12.3-py39hf3d152e_7 --> 5.12.3-py39hf3d152e_8
  pyqt-impl                           5.12.3-py39h0fcd23e_7 --> 5.12.3-py39hde8b62d_8
  pyqt5-sip                          4.19.18-py39he80948d_7 --> 4.19.18-py39he80948d_8
  pyqtchart                             5.12-py39h0fcd23e_7 --> 5.12-py39h0fcd23e_8
  pyqtwebengine                       5.12.1-py39h0fcd23e_7 --> 5.12.1-py39h0fcd23e_8
  pysocks                              1.7.1-py39hf3d152e_3 --> 1.7.1-py39hf3d152e_4
  python                           3.9.6-h49503c6_1_cpython --> 3.9.10-h85951f9_1_cpython
  pytz                                  2021.1-pyhd8ed1ab_0 --> 2021.3-pyhd8ed1ab_0
  pywavelets                           1.1.1-py39hce5d2b2_3 --> 1.2.0-py39hce5d2b2_1
  qt                                      5.12.9-hda022c4_4 --> 5.12.9-ha98a1a1_5
  rasterio                             1.2.6-py39hbc4e497_2 --> 1.2.10-py39h0401cea_4
  requests                              2.26.0-pyhd8ed1ab_0 --> 2.27.1-pyhd8ed1ab_0
  setuptools                          49.6.0-py39hf3d152e_3 --> 60.6.0-py39hf3d152e_0
  shapely                              1.7.1-py39ha61afbd_5 --> 1.8.0-py39ha65c37e_5
  snowballstemmer                        2.1.0-pyhd8ed1ab_0 --> 2.2.0-pyhd8ed1ab_0
  sphinx                                 4.1.2-pyh6c4a22f_1 --> 4.4.0-pyh6c4a22f_1
  sphinxcontrib-ser~                     1.1.5-pyhd8ed1ab_0 --> 1.1.5-pyhd8ed1ab_1
```

Both documentation and tests pass fine on my machine. Let's try an empty commit and see how tests do. I'd like them to pass on the existing 3.9 environment in CI before updating to 3.10.

djhoese:
So I wasn't able to reproduce the issue locally, but then I realized that the conda packages being installed on CI were for Python 3.10. The conda environment was never being told what version of python to use so it was using the latest (Python 3.10), even though we thought we were testing 3.7 and 3.9. I'm updated the CI so it should be using 3.9 now. If that passes I will then build a local 3.10 environment and try to reproduce the issues we were seeing here in CI. I will then also update CI to run Python 3.7 and 3.10.

djhoese:
Ok, this looks better. The failing test is expected as the current version of numba does not support the newest version of numpy and fails to import when it checks for that.

djhoese:
@rayg-ssec So I think I've fixed everything now. The unstable test environment (last parameter is `true`) will remain failing until a new version of numba is released. I'm not sure of a good way around that.

As for the PyQt5 issues, I've described the issue in this vispy issue here:

https://github.com/vispy/vispy/issues/2212#issuecomment-1026993206

Bottom line is this seems to be a PyQt5 bug that, based on changelog entries, has been fixed in at least PyQt6. The URL for the changelog for PyQt5 is broken so I can't be sure if later versions of PyQt5 will have this fixed. It is not an issue with all `setAlignment` functions, only the `QGroupBox` from what I can tell. In Python 3.10, low-level implicit integer conversion is no longer done. In PyQt5 the AlignmentFlag object can be converted to an integer just fine, but the Alignment object resulting from `|`'ing of the AlignmentFlag objects does not appear to be able to handle this. The simplest workaround is to modify pyuic5 generated python files with an `int` wrapping the flag usage in `setAlignment` *only* for `QGroupBox` widgets.

doc/source/conf.py
```diff
@@ -54,6 +54,23 @@
...
+    if what == "class" and "pyqtSignal" in obj.__class__.__name__:
+        return True
+    # fallback to other handlers or autodoc default
+    return None
+
+
+def setup(app):
+    """Global setup function for sphinx documentation.
+
+    Connect to autodoc event so invalid pyqtSignal docs can be skipped.
```
estan, line 68: We ran across this in our project as well, and found your workaround here.

But I'm curious, what is it in `pyqtSignal`'s docstring that is invalid? I'm staring at it, comparing it to e.g. that for `pyqtSlot`, but I can't see what's special about it :p

djhoese: Oh, it's been a while and I don't quite remember exactly. I think on certain versions of Python and certain versions of PyQt, the pyqtSignal's `__doc__` property was set to something weird or caused a segfault or something when it was read. We, as users of pyqtsignal, can't set the docstring through normal means as far as I know so this was necessary. We could make a PR and try removing this skip logic and see what happens. Let me try it locally first...

estan: Ah, but I confirmed that your workaround works for us. We ran into the problem on Python 3.10, using 5.15.2.

But I'm print the `__doc__` of `pyqtSignal`, and I can't see the reason why sphinx is complaining:

```
(env) orexplore@vidar-m7:~$ python3 -c 'from PyQt5.QtCore import pyqtSignal; print(pyqtSignal.__doc__)'
pyqtSignal(*types, name: str = ..., revision: int = ..., arguments: Sequence = ...) -> PYQT_SIGNAL

types is normally a sequence of individual types.  Each type is either a
type object or a string that is the name of a C++ type.  Alternatively
each type could itself be a sequence of types each describing a different
overloaded signal.
name is the optional C++ name of the signal.  If it is not specified then
the name of the class attribute that is bound to the signal is used.
revision is the optional revision of the signal that is exported to QML.
If it is not specified then 0 is used.
arguments is the optional sequence of the names of the signal's arguments.

(env) orexplore@vidar-m7:~$
```

djhoese: Ok, got it. So this was added because when sphinx docs are generated it produces a ton of these warnings for each pyqtsignal:

```
docstring of uwsift.queue.TaskQueue.didMakeProgress:13: WARNING: Inline emphasis start-string without end-string.
```

This is what the docstring for all pyqtsignals looks like:

```python
In [6]: Document.didAddBasicLayer.__doc__
Out[6]: "pyqtSignal(*types, name: str = ..., revision: int = ..., arguments: Sequence = ...) -> PYQT_SIGNAL\n\ntypes is normally a sequence of individual types.  Each type is either a\ntype object or a string that is the name of a C++ type.  Alternatively\neach type could itself be a sequence of types each describing a different\noverloaded signal.\nname is the optional C++ name of the signal.  If it is not specified then\nthe name of the class attribute that is bound to the signal is used.\nrevision is the optional revision of the signal that is exported to QML.\nIf it is not specified then 0 is used.\narguments is the optional sequence of the names of the signal's arguments.\n"
```

I don't like generating sphinx docs that contain warnings so I typically add `-W` to make it fail when it encounters warnings.

I just saw your new comment...one sec...

djhoese: It is the `*` in `*types` that in RST starts a bolded text, but it is never enclosed.

I chose this simple workaround rather than forcing myself to modify ever pyqtsignal object and all future signals. I was hoping for something easier or better, but couldn't come up with it at the time.

estan: Alright, yea I was just wondering why we're seeing this for pyqtSignal and not pyqtSlot, which we're also using and which also has an asterisk there.

estan: (same here we treat warnings as errors and that's how we discovered the issue. It only appears on Python 3.10, not on 3.8 which we also test)

estan: I'll doublecheck what the docstring looks like for our pyqtSlot-decorated functions, perhaps the pyqtSlot docstring doesn't propagate the same way the pyqtSignal one does.

djhoese: That or wrap it in double backticks to represent a literal.

PR #335, opened 2022-05-13
Merge features, fixes, and rewrites performed by EUMETSAT

djhoese:
This is a giant PR that merges everything from the EUMETSAT upstream `master` branch. This is 2+ years of work I think. The merge conflicts were pretty simple, but overall this is a lot of files.

Conflicts:

1. Import changes in workspace.py. We made updates for Python 3.10 (import from collections.abc) and they just overall changed imports to fit their changes.
2. Two `_ui.py` files where we had to make hacks to add an `int()` call to some Enum usage due to a bug in PyQt. This was a hack then and is a hack now. It should be retested with newer versions of PyQt from conda-forge. If it still fails with that then it should be done with a newer version on PyPI if possible.

As mentioned in my "black" PR to the EUMETSAT gitlab, there are a lot of flake8 issues from EUMETSAT's master branch. I'll need to fix those before this is merged and then push the result of all of that to a EUMETSAT gitlab merge request (we'll see how that goes).

FYI to @sjoro


djhoese:
So I've fixed all the flake8 issues and the docstring issues that were stopping the docs from generating without warnings...but we have a problem.

It seems the tests haven't been run on the EUMETSAT branch since September 2020 at least. There are tests in `test_algebraic.py` that import `Workspace`, but that class was refactored in September 2022 into `BaseWorkspace`/`CachingWorkspace`. That import is an easy fix, but there are many other tests that are failing.

djhoese:
This branch was merged into EUMETSAT's gitlab some time ago. It no longer reflects the current state of EUM's version so I'm closing it.

PR #363, opened 2023-05-23
Migrate contributing and test writing docs from wiki to sphinx

djhoese:
This PR migrates the contributor's guide from the Wiki as well as the documentation on writing tests and puts them in the sphinx docs under a new Developer's Guide.

I think I'd still like to migrate the installation instructions just to make sure the language between what was on the wiki and what was put in the sphinx docs are similar and cover the same use cases.


djhoese:
Once this is merged I will update the contributing wiki page to point to the RTD site (if not remove the wiki page completely).

doc/source/installation.rst
```diff
@@ -0,0 +1,172 @@
...
+^^^^^^^^^^^^^^
+
+SIFT is distributed as a compressed `.zip` archive with all dependencies
+provided internally. The bundle (.zip) can be downloaded from our FTP server
+`here <https://bin.ssec.wisc.edu/pub/sift/dist/>`_. Once downloaded and the zip
+file extracted, the file `SIFT.bat` can be double clicked to start SIFT.
+
+By default SIFT caches files in a "workspace" located
+at the user's
+``/Users/<User>/AppData/Local/CIMSS-SSEC/SIFT/Cache/workspace`` directory.
```
ameraner, line 52:
```suggestion
``\Users\<User>\AppData\Local\SIFT\Cache\workspace`` directory.
```

doc/source/installation.rst
```diff
@@ -0,0 +1,172 @@
...
+
+.. code-block:: bash
+
+   tar -xzf SIFT_1.1.2.tar.gz
+
+This will create a ``SIFT`` directory. SIFT can then be run by doing:
+
+.. code-block:: bash
+
+   SIFT/SIFT.sh
```
ameraner, line 72:
```suggestion
   tar -xzf SIFT_X.Y.Z.tar.gz

This will create a ``SIFT_X.Y.Z`` directory. SIFT can then be run by doing:

.. code-block:: bash

   SIFT_X.Y.Z/SIFT.sh
```

ameraner (changes requested):
Thank you for this! Please see the comments/suggestions I left.

doc/source/installation.rst
```diff
@@ -0,0 +1,172 @@
...
+	 this channel, by default it is ``~/conda-channels/uwsift/`` (see
+	 :ref:`conda-packaging`)
+
+For Users
+^^^^^^^^^
+
+Install the package *uwsift* into an environment called e.g. ``work`` and
+prepared as described above::
+
+  (work)%> conda install uwsift
```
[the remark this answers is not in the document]
ameraner: Oh, I see now that there's already a nicer conda installation guidance here: https://github.com/ssec/sift/wiki/installation-guide#python-package-installation, so maybe we can just bring this over?

ameraner:
Regarding your point on the developer installation and the requirements. I agree, having a visible `requirements.yaml` for this purpose would be nice! In principle it would be the same as the requirements in `setup.py`, no? Is it possible to have a unified yaml file and point `setup.py` to it?

doc/source/dev_guide/contributing.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Starting with SIFT 1.1 Satpy is used for reading all input data files. By
+default SIFT limits the available readers to avoid unsupported data formats
+and other unexpected behavior. To customize the readers used you can specify
+the :envvar:`UWSIFT_DATA_READING__READERS` environment variables (note the last
+specified has two underscores). You can set it to a list of readers. For example:
+
+.. code-block:: bash
+
+   export UWSIFT_DATA_READING_READERS = "['abi_l1b', 'ami_l1b']"
```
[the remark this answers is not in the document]
djhoese:
```suggestion
Starting with SIFT 1.1 Satpy is used for reading all input data files. 
Check :doc:`the reader configuration docs </configuration/readers>` to configure which Satpy 
readers are made available in SIFT. If you have a data source that is not yet supported in Satpy 
(available readers are listed `here <https://satpy.readthedocs.io/en/stable/index.html#id1>`_ ), you 
have to write a Satpy reader in order to be able to visualize it in SIFT. Head to the Satpy 
documentation for guidelines on writing a new reader `here <https://satpy.readthedocs.io/en/stable/dev_guide/custom_reader.html>`_, and consider 
contributing your reader to Satpy via a GitHub Pull Request. If you don't want to publish your reader, or as an 
intermediate solution, you can plugin "local" readers in SIFT, see the :doc:`external satpy documentation </configuration/external_satpy>`
```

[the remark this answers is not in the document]
djhoese: I changed a little bit of the wording in your suggestion and changed `visualise` to the US spelling `visualize`.

doc/source/installation.rst
```diff
@@ -0,0 +1,172 @@
...
+System Requirements
+-------------------
+
+SIFT works on Windows, Mac, and Linux, but depends on many complex libraries
+and relies heavily on GPU (video card) processing via the PyOpenGL and VisPy
+libraries which may make installation difficult on some platforms or systems.
+In general it is best to have a local physical system to run SIFT as opposed
+to a virtual machine or remote server. Virtual machines and remote servers
+commonly have limited ways of displaying OpenGL visualizations to another
+machine.
```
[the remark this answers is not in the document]
djhoese: I would say "relies heavily on GPU" is a bit too strong of a statement, no? It seems to me that it works ok also on simpler systems without it. I'm just afraid that this may scare off potential users that have limited resources. 
```suggestion
SIFT works on Windows, Mac, and Linux, but depends on many complex
libraries which may make installation difficult on some platforms or systems.
It also makes heavy use of GPU (video card) processing via the PyOpenGL and VisPy
libraries. In general it is best to have a local physical system to run SIFT as opposed
to a virtual machine or remote server. Virtual machines and remote servers
commonly have limited ways of displaying OpenGL visualizations to another
machine.
```

doc/source/installation.rst
```diff
@@ -0,0 +1,172 @@
...
+
+While the below system resources should work they should be considered the
+lower end of machines for running SIFT. In general the better GPU and faster
+memory (RAM) the better experience you'll have with SIFT.
+
+* Windows 10+ / Mac OS X >11.0 / Linux >= Rocky Linux 8
+* 16GB RAM
+* SSD/NVMe drive with 128GB+ available
+* GPU with 4GB VRAM and OpenGL 3+ support
+* Data files to be loaded can require up to 1TB or more
```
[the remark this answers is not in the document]
djhoese: I'm a little worried about the 2GB VRAM, but ok.

doc/source/installation.rst
djhoese: I'm not sure this would render properly. I think this was written with `/` is because Python understands it, but I get your point.

doc/source/installation.rst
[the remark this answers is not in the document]
djhoese: Yes, that was going to be my next task.

djhoese:
> Regarding your point on the developer installation and the requirements. I agree, having a visible requirements.yaml for this purpose would be nice! In principle it would be the same as the requirements in setup.py, no? Is it possible to have a unified yaml file and point setup.py to it?

No. The YAML is used by conda, setup.py is only used by pip. There is a chance package names are different and we're actually running into that with the readthedocs stuff where if you do `pip install -e .` in a conda environment that already has `pyqt=5.x` installed it will try to install `pyqt5` from PyPI because the package names are different.

doc/source/dev_guide/contributing.rst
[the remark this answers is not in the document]
ameraner: Sounds good, thanks!

doc/source/installation.rst
ameraner: Yeah, I think the most frequent use case here is probably the user copy-pasting (part of) the path in the Explorer, so it's just nice if it works then

doc/source/installation.rst
```diff
@@ -0,0 +1,115 @@
...
+machine.
+
+While the below system resources should work they should be considered the
+lower end of machines for running SIFT. In general the better GPU and faster
+memory (RAM) the better experience you'll have with SIFT.
+
+* Windows 10+ / Mac OS X >11.0 / Linux >= Rocky Linux 8
+* 8GB RAM
+* Disk space (preferrably on a SSD/NVMe drive) with 20GB+ available
+* GPU with 2GB VRAM and OpenGL 3+ support
```
ameraner, line 30: Sorry to reiterate on this. From my last demos at various conferences, one of the first question is always: how will it perform on a simple laptop, without a GPU? My answer was usually: it's ok, as long as you don't load too many layers/time steps/large-array-datasets, as this is my experience on our EUM laptops. I'm just worried that someone looking for the specs and reading about the heavy GPU usage, might decide to not even try if she/he doesn't have one. 

How about pointing out that GPUs are optional, but highly recommended?
```suggestion
* GPU with 2GB VRAM and OpenGL 3+ support (optional, but highly recommended)
```

ameraner (approved):
Just a final comment on the GPU specs, otherwise LGTM, thanks a lot again!

doc/source/dev_guide/contributing.rst
```diff
@@ -0,0 +1,201 @@
...
+
+Additional Satpy Readers
+------------------------
+
+Starting with SIFT 1.1 Satpy is used for reading all input data files. 
+Check :doc:`the reader configuration docs </configuration/readers>` to configure which Satpy 
+readers are made available in SIFT. If you have a data source that is not yet supported in Satpy 
+(available readers are listed `here <https://satpy.readthedocs.io/en/stable/index.html#id1>`_ ), you 
+have to write a Satpy reader in order to be able to visualize it in SIFT. Head to the Satpy 
+documentation for guidelines on writing a new reader `here <https://satpy.readthedocs.io/en/stable/dev_guide/custom_reader.html>`_, and consider
```
ameraner, line 161: On, the Sphinx build is failing now
```
Warning, treated as error:
/home/runner/work/sift/sift/doc/source/dev_guide/contributing.rst:2:Duplicate explicit target name: "here".
```
I think this should fix it? (https://github.com/sphinx-doc/sphinx/issues/3921#issuecomment-315581557)
```suggestion
(available readers are listed `here <https://satpy.readthedocs.io/en/stable/index.html#id1>`__ ), you 
have to write a Satpy reader in order to be able to visualize it in SIFT. Head to the Satpy 
documentation for guidelines on writing a new reader `here <https://satpy.readthedocs.io/en/stable/dev_guide/custom_reader.html>`__, and consider 
```

djhoese: Oh I've never seen that anonymous references idea. That's awesome. I always reworded things so I wasn't using `here`. Great!

doc/source/installation.rst
djhoese: I disagree. Users don't need a monster of a machine and that's fine that we lowered these limits, but we're talking about some high resolution satellite data and you've changed the default image behavior to require loading the entire image into GPU memory. Regardless of what the results of testing were (tiling didn't release memory, etc), this at least logically means that more GPU memory is going to be used. Telling people they don't need a good GPU or that an Intel CPU-integrated GPU is good enough is really risky in my opinion.

Users purchase machines based on these specs. I don't care if someone says "let's see if I can get it to run on my system" for a system they already have and that it works fine on their low-end system. They were going to do that testing anyway. I don't want users to buy new machines, especially if buying multiple for a training course, and have none of those machines work well.

Overall we should do more testing of how things really perform on lower end hardware.

djhoese:
pre-commit.ci autofix

doc/source/installation.rst
ameraner: Ok, I trust your opinion then :)

doc/source/dev_guide/contributing.rst
```diff
@@ -0,0 +1,202 @@
...
+-------------------
+
+If you have an idea for a feature you'd like to implement in SIFT please create
+an issue on GitHub to discuss the feature with the core developers. You can
+also chat with SIFT developers on the
+`Gitter chat <https://gitter.im/ssec/sift>`_. They may be able to help guide
+you on how to add the feature.
+
+.. _dev_install:
+Developer Installation
```
djhoese, line 73:
```suggestion
.. _dev_install:

Developer Installation
```

djhoese:
I migrated the installation instructions from the wiki. I think this is good to go for now.

PR #364, opened 2023-05-24
Implement DYNAMIC kind to handle ambiguous readers like the LI reader

ameraner:
The PR attempts to handle the case where readers can return multiple types of data, on the example of the `li_l2_nc` reader that can return both gridded image data, as well as point data.


uwsift/workspace/importer.py
```diff
@@ -715,6 +715,12 @@ def _set_kind_metadata(self, attrs: dict) -> None:
             LOG.info(f"No data kind configured for reader '{self.reader}'." f" Falling back to 'IMAGE'.")
             attrs[Info.KIND] = Kind.IMAGE
 
+        if attrs[Info.KIND] is Kind["IMAGEorPOINTS"]:
+            if isinstance(attrs["area"], SwathDefinition) and len(attrs["area"].shape) == 1:
+                attrs[Info.KIND] = Kind.POINTS
+            else:
+                attrs[Info.KIND] = Kind.IMAGE
```
djhoese, line 722: I was thinking/wondering if maybe the "image or points" kind wasn't an actual kind, but more of a special config-only value that is interpreted in this method (or some new one that makes it more clear that it is doing this conversion, outside the importer). So basically this method in the importer always assigns a real `Kind` enum, but the input "kind" from the config could be other special values like "dynamic" or "image_or_points", but those don't actually exist in the `Kind` enum.

Does that make sense?

My worry is that we don't want to add more to `Kind` because we don't want the other parts of SIFT to have to handle that.

Additionally, I think I like the idea of a "dynamic" value rather than a specific "this can only be an image or points" data. The dynamic at-load time behavior would be the same for most if not all cases. It also gets around readers adding new types of data in the future. Thoughts?

ameraner: Good points, thanks for the feedback!

I also thought about the fully-dynamic approach, and I agree it is more elegant (there's the risk we end up with too many permutations), but my initial worry was that we don't have, at the moment, enough metadata in the `attrs` to be able to decide dynamically every time (e.g. between other types combinations). On the other hand, we could implement `DYNAMIC`, and document that atm it handles only the image/point ambiguity with the given assumptions, and more combinations will be supported in the future if needed - this is what I now did.

I refactored the code to not add a kind anymore as well.

uwsift/workspace/importer.py
```diff
@@ -705,15 +729,14 @@ def _set_kind_metadata(self, attrs: dict) -> None:
         if "prerequisites" in attrs:
             attrs[Info.KIND] = Kind.MC_IMAGE
             return
+
         reader_kind = config.get(f"data_reading.{self.reader}.kind", None)
```
djhoese, line 733: Maybe in a future PR this default should be `"DYNAMIC"?

djhoese (approved):
I think this looks pretty good. I think this leaves us some flexibility in the future for making this all automated so it doesn't require any configuration. Similarly, it'd be nice if the resampling could "just work" with logical defaults and not need readers to be configured beforehand. It would have to be a similar thing where datasets are resampled a certain way depending on there type. For example, SIFT detects the Scene contains 2D swath defs, 1D swath defs, and area defs. It could try native resampling for just the areas if it detects they're the same, do nearest neighbor by default for both types of swath defs, and so on.

Issue #380, opened 2023-07-11
Allow setting of satpy config path via env variable


nedelceo:
At this moment setting extra [Satpy component configuration path](https://satpy.readthedocs.io/en/stable/config.html#component-configuration-path) is done by setting `satpy_extra_readers_import_path` in [external_satpy_and_readers.yaml](https://github.com/ssec/sift/blob/6ca36f789fce67d34affb05fa101ee026f623a00/uwsift/etc/SIFT/config/external_satpy_and_readers.yaml). After discussion on [gitter](https://matrix.to/#/!AXWZGEQAHtBDfySPVa:gitter.im/$zJJMG5dFJIh-WkI26SLAqUW7gWZ638jQwnjZ5AML7UE?via=gitter.im), I think this could be improved in two steps.

1) Rename `satpy_extra_readers_import_path` because now  the name is confusing. Since it it is not only related to readers definitions, but also to definitions of composites, enhancements, areas and so on. The variable could be renamed to e.g. `satpy_extra_config_path`.

2) Allow the use of environment variable to define satpy config path. Now environment variable `SATPY_CONFIG_PATH` is set to  `satpy_extra_readers_import_path`   https://github.com/ssec/sift/blob/6ca36f789fce67d34affb05fa101ee026f623a00/uwsift/__init__.py#L90-L93
 I think it would be useful not to reset  `SATPY_CONFIG_PATH` but append `satpy_extra_readers_import_path` to it  if this env variable is already defined.

djhoese:
Sorry about not contributing to the discussion on gitter. I've been having a really hard time getting gitter to notify me at all.

I agree with all of this. I believe newer versions of `donfig` (the package I maintain and that we use for the configuration logic here) allow us to specify deprecated or renamed config entries. If anyone has time to put together a pull request these would be great changes.

One thing I'm thinking about the `satpy_extra_config_path`...I'm wondering if this is needed at all. I mean, the main reason for `satpy_extra_readers_import_path` is so that SIFT adds it as an importable location for Python. SIFT could just do this for all `SATPY_CONFIG_PATH` directories. It is a little risky security-wise. Otherwise though, this configuration value is a Satpy value, not a SIFT value. Would it be too complex to expect users to understand that they are configuring Satpy at this point and to specify a separate Satpy configuration file and/or environment variables specific to Satpy? I mean, these users would be customizing Satpy, so that seems like a reasonable assumption to me.

djhoese:
[...]

PR #381, opened 2023-07-13
Fix setting of external Satpy component configuration

nedelceo:
Setting external Satpy component configuration is done by setting `satpy_extra_readers_import_path` which is missleading as described in  #380  

This PR:
1. renames  `setting satpy_extra_readers_import_path`  to `satpy_extra_config_path`
2. allows setting of satpy config path via env variable ` SATPY_CONFIG_PATH`

### DIFF
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -26,7 +26,7 @@ one providing Satpy, the application may exit immediately throwing Exceptions.
 Using Extra Satpy component configuration
 =========================================
 
-SIFT can use external `Satpy component configuration <https://satpy.readthedocs.io/en/stable/config.html#component-configuration>`_ folder,
+SIFT can use external :ref:`Satpy component configuration <satpy:component_configuration>` folder,
 that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
 To use external satpy component configuration it is necessary to define
 ``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_::
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -24,7 +24,7 @@ package: If the given path doesn't point to a Python package directory or not to
 one providing Satpy, the application may exit immediately throwing Exceptions.
 
 Using Extra Satpy component configuration
-===================
+=========================================
 
 SIFT can use external `Satpy component configuration <https://satpy.readthedocs.io/en/stable/config.html#component-configuration>`_ folder,
 that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
@@ -36,7 +36,7 @@ To use external satpy component configuration it is necessary to define
 or it can be defined via environmet variable ``SATPY_CONFIG_PATH`` as described `here <https://satpy.readthedocs.io/en/stable/config.html#config-path-setting>`_.
 
 Example of external readers configuration
-`````````````
+`````````````````````````````````````````
 
 Several data formats which are or will be produced by EUMETSAT need special
 readers which are not (yet) part of the official Satpy distribution. EUMETSAT
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -23,18 +23,26 @@ It is your responsibility to make sure the setting points to a suitable Satpy
 package: If the given path doesn't point to a Python package directory or not to
 one providing Satpy, the application may exit immediately throwing Exceptions.
 
-Using Extra Readers
+Using Extra Satpy component configuration
 ===================
+
+SIFT can use external `Satpy component configuration <https://satpy.readthedocs.io/en/stable/config.html#component-configuration>`_ folder,
+that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
+To use external satpy component configuration it is necessary to define
+``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_: ::
+
+    satpy_extra_config_path: [directory path]
+
+or it can be defined via environmet variable ``SATPY_CONFIG_PATH`` as described `here <https://satpy.readthedocs.io/en/stable/config.html#config-path-setting>`_.
+
+Example of external readers configuration
+`````````````
 
 Several data formats which are or will be produced by EUMETSAT need special
 readers which are not (yet) part of the official Satpy distribution. EUMETSAT
 maintains a Git repository ``satpy/local_readers`` on their `GitLab
 <https://gitlab.eumetsat.int/satpy/local_readers>`_ providing these special
-readers. To use these readers in addition to those included in Satpy the path to
-the root directory of a clone of this repository must be configured via the
-following setting first::
-
-    satpy_extra_readers_import_path: [directory path]
+readers. To use these readers it is neccsary to put them into folder: ``satpy_extra_config_path/readers``.
 
 Furthermore the desired readers need to be added to the configuration
 ``data_reading.readers`` and their reader specific configuration as well (see
@@ -42,13 +50,13 @@ Furthermore the desired readers need to be added to the configuration
 
 For example assuming that the repository has been cloned as follows::
 
-    git clone https://gitlab.eumetsat.int/satpy/local_readers.git /path/to/clone/of/satpy/local_readers
-
+    git clone https://gitlab.eumetsat.int/satpy/local_readers.git /path/to/satpy_extra_config_path/readers
+    
 the readers for the *FCI L1 Landmark Locations Catalogue*, *FCI L1 GEOOBS
 Landmarks* (landmark locations) and *FCI L1 GEOOBS Landmark Matching Results*
-(landmark navigation error) can be made available in SIFT with::
+(landmark navigation error) can be made available in MTG-SIFT with::
 
-    satpy_extra_readers_import_path: /path/to/clone/of/satpy/local_readers
+    satpy_extra_config_path: /path/to/satpy_extra_config_path
 
     data_reading:
       readers:
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -54,7 +54,7 @@ For example assuming that the repository has been cloned as follows::
 
 the readers for the *FCI L1 Landmark Locations Catalogue*, *FCI L1 GEOOBS
 Landmarks* (landmark locations) and *FCI L1 GEOOBS Landmark Matching Results*
-(landmark navigation error) can be made available in MTG-SIFT with::
+(landmark navigation error) can be made available in SIFT with::
 
     satpy_extra_config_path: /path/to/satpy_extra_config_path
 
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -28,7 +28,7 @@ Using Extra Satpy component configuration
 
 SIFT can use external :ref:`Satpy component configuration <satpy:component_configuration>` folder,
 that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
-To use the external satpy component configuration it is necessary to define either 
+To use the external satpy component configuration it is necessary to define either
 ``satpy_extra_config_path`` in the personal user configs (e.g. inside a file called `external_satpy.yaml`)::
 
     satpy_extra_config_path: [directory path]
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -51,7 +51,7 @@ Furthermore the desired readers need to be added to the configuration
 For example assuming that the repository has been cloned as follows::
 
     git clone https://gitlab.eumetsat.int/satpy/local_readers.git /path/to/satpy_extra_config_path/readers
-    
+
 the readers for the *FCI L1 Landmark Locations Catalogue*, *FCI L1 GEOOBS
 Landmarks* (landmark locations) and *FCI L1 GEOOBS Landmark Matching Results*
 (landmark navigation error) can be made available in MTG-SIFT with::
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -29,7 +29,7 @@ Using Extra Satpy component configuration
 SIFT can use external `Satpy component configuration <https://satpy.readthedocs.io/en/stable/config.html#component-configuration>`_ folder,
 that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
 To use external satpy component configuration it is necessary to define
-``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_: ::
+``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_::
 
     satpy_extra_config_path: [directory path]
 
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -29,7 +29,8 @@ Using Extra Satpy component configuration
 SIFT can use external :ref:`Satpy component configuration <satpy:component_configuration>` folder,
 that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
 To use external satpy component configuration it is necessary to define
-``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_::
+``satpy_extra_config_path`` in ``*.yaml`` file in USER_SIFT_CONFIG_DIR according to the example in 
+`external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_::
 
     satpy_extra_config_path: [directory path]
 
--- a/doc/source/configuration/external_satpy.rst
+++ b/doc/source/configuration/external_satpy.rst
@@ -1,30 +1,5 @@
 Configuring External Satpy Components
--------------------------------------
-
-Replacing Satpy by External Installation
-========================================
-
-SIFT can be instructed to import Satpy modules from another location than
-from the site packages of the active Python environment when the following
-setting points to an appropriate package directory::
-
-   satpy_import_path: [directory path]
-
-For example you can use your development version of Satpy cloned directly from
-GitHub to ``/home/me/development/satpy`` by configuring::
-
-   satpy_import_path: "/home/me/development/satpy/satpy"
-
-or setting the according environment variable before starting SIFT::
-
-   export UWSIFT_SATPY_IMPORT_PATH="/home/me/development/satpy/satpy"
-
-It is your responsibility to make sure the setting points to a suitable Satpy
-package: If the given path doesn't point to a Python package directory or not to
-one providing Satpy, the application may exit immediately throwing Exceptions.
-
-Using Extra Satpy component configuration
-=========================================
+=====================================
 
 SIFT can use external :ref:`Satpy component configuration <satpy:component_configuration>` folder,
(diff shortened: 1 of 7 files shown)


nedelceo:
I am not sure how to test changes. Should I create similar to `test_satpy_import.py`

doc/source/configuration/external_satpy.rst
```diff
@@ -23,32 +23,40 @@ It is your responsibility to make sure the setting points to a suitable Satpy
...
 
-Using Extra Readers
-===================
+Using Extra Satpy component configuration
+=========================================
+
+SIFT can use external `Satpy component configuration <https://satpy.readthedocs.io/en/stable/config.html#component-configuration>`_ folder,
+that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
+To use external satpy component configuration it is necessary to define
+``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_: ::
```
djhoese, line 32:
```suggestion
``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_::
```

djhoese: This should probably refer to where this file shows up in the final installation too, right?

doc/source/configuration/external_satpy.rst
```diff
@@ -23,32 +23,40 @@ It is your responsibility to make sure the setting points to a suitable Satpy
...
 
 For example assuming that the repository has been cloned as follows::
 
-    git clone https://gitlab.eumetsat.int/satpy/local_readers.git /path/to/clone/of/satpy/local_readers
+    git clone https://gitlab.eumetsat.int/satpy/local_readers.git /path/to/satpy_extra_config_path/readers
 
 the readers for the *FCI L1 Landmark Locations Catalogue*, *FCI L1 GEOOBS
 Landmarks* (landmark locations) and *FCI L1 GEOOBS Landmark Matching Results*
-(landmark navigation error) can be made available in SIFT with::
+(landmark navigation error) can be made available in MTG-SIFT with::
```
djhoese, line 57:
```suggestion
(landmark navigation error) can be made available in SIFT with::
```

Is there a reason you added MTG back in? This was intentionally removed from past versions of the documentation.

uwsift/__init__.py
```diff
@@ -87,10 +87,20 @@ def find_spec(cls, fullname: str, path=None, target=None):
...
-    os.environ["SATPY_CONFIG_PATH"] = satpy_extra_readers_import_path
+
+# get satpy_config_path from config file
+satpy_config_path_yml = config.get("satpy_extra_config_path", None)
+# get satpy_config_path from env variable
+satpy_config_path_env = os.getenv("SATPY_CONFIG_PATH", None)
+# expand or set environment variable SATPY_CONFIG_PATH if satpy_config_path is defined in configuration file
+if satpy_config_path_yml is not None:
+    if satpy_config_path_env is not None:
+        os.environ["SATPY_CONFIG_PATH"] = satpy_config_path_env.replace('"', "") + os.pathsep + satpy_config_path_yml
```
djhoese, line 98: What is the string replacing for? Do environment variables usually have quotes in them?

djhoese: And can we simplify this whole logic by making it two if statements:

```
if satpy_config_path_yml is not None:
    satpy_config_path_env = os.getenv("SATPY_CONFIG_PATH", "")
    if satpy_config_path_env:
        satpy_config_path_env += os.pathsep
    os.environ["SATPY_CONFIG_PATH"] = satpy_config_path_env + satpy_config_path_yml
```

Also, this seems a little risky in the long run. This depends on Satpy not being imported before this code runs. That might work now, but may not be true in the future. I think it'd be more future proof to import Satpy explicitly and then do 

```python
satpy.config.set(config_path=satpy.config.get("config_path") + [satpy_config_path_yml])
```

djhoese (changes requested):
Thanks for starting this. I had a couple comments. Let me know what you think.

djhoese:
About tests...it may be hard to test this since everything happens at import time. If we moved the main part of the logic into a function you could probably run the function in your test and make sure it changes the satpy config with different sift config values.

uwsift/__init__.py
nedelceo: > What is the string replacing for? Do environment variables usually have quotes in them?

I worked with an example `set SATPY_CONFIG_PATH="C:/path/custom1;C:/path/custom2"` from [Satpy docs](https://satpy.readthedocs.io/en/stable/config.html#component-configuration) .

nedelceo: Thank you for suggesting simpler and more future proof solution. I committed this version, that works , if there is not the windows quotes problem involved. 

```
# get satpy_config_path from config file
satpy_config_path_yml = config.get("satpy_extra_config_path", None)
if satpy_config_path_yml is not None:
    satpy.config.set(config_path=satpy.config.get("config_path") + [satpy_config_path_yml])
```

doc/source/configuration/external_satpy.rst
```diff
@@ -23,32 +23,40 @@ It is your responsibility to make sure the setting points to a suitable Satpy
 package: If the given path doesn't point to a Python package directory or not to
 one providing Satpy, the application may exit immediately throwing Exceptions.
 
-Using Extra Readers
-===================
+Using Extra Satpy component configuration
+=========================================
+
+SIFT can use external `Satpy component configuration <https://satpy.readthedocs.io/en/stable/config.html#component-configuration>`_ folder,
```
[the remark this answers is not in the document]
nedelceo: this change does not work in my local repositroy. When I `make html` I got `D:\git\sift_fork\doc\source\configuration\external_satpy.rst:29: WARNING: undefined label: 'satpy:component-configuration'` and I was not able to resolve it.

doc/source/configuration/external_satpy.rst
djhoese: Sorry, I meant maybe you could also mention (in addition to linking to the github version of the YAML) that in the final installation (when someone installs SIFT) that this file would show up in `SIFT_directory/lib/python3.11/site-packages/uwsift/etc/...` or wherever it ends up.

doc/source/configuration/external_satpy.rst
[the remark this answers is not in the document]
djhoese: Before giving up on it, could you try `component_configuration` (with an underscore) and see if that works better? I've had mixed success using references like this between projects with intersphinx. I think in one case I ended up just doing the URL as you have it here because intersphinx would not find the reference no matter what I did.

doc/source/configuration/external_satpy.rst
ameraner: Actually, as described here https://sift.readthedocs.io/en/latest/configuration/index.html#configuration, the idea is that users do not modify the default configuration (e.g. in `SIFT_directory/lib/python3.11/site-packages/uwsift/etc/`), but rather place the personal settings in the user config path (e.g. `~/.config/SIFT/config/`), so that in the case of an update, the configuration modifications are not lost. So I would refrain in pointing out where the defaults are, if you agree. 

Also, technically the user config doesn't need to be in a file with a specific name; the file just needs to be a .yaml file inside the user config folder. So I would also not mention `external_satpy.yaml` in the documentation, but rather simply say `in the user configs`

djhoese: Ah right. Ok but will `external_satpy.yaml` show up in a users config directory by default? No, right? So this `external_satpy.yaml` is just an example. So this section of the docs should maybe point the user to their personal configuration directory and say that `external_satpy.yaml` is an example?

ameraner: No it won't, the folder is empty (apart from the empty subfolder `readers`). Yep, agree.

ameraner (changes requested):
Thank you for your work on this, I agree the naming and the setup needed to be improved! 
I left some comments and proposed some changes.

doc/source/configuration/external_satpy.rst
ameraner: No it won't, the folder is empty (apart from the empty subfolder `readers`). Yep, agree.

doc/source/configuration/external_satpy.rst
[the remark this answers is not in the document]
nedelceo: That was it, the underscore. Thanks.

doc/source/configuration/external_satpy.rst
nedelceo: Ok, I made f7287c405c5490d80b1abde391a9b46e1f4f02a4

doc/source/configuration/external_satpy.rst
```diff
@@ -23,32 +23,40 @@ It is your responsibility to make sure the setting points to a suitable Satpy
...
+that hosts extra ``readers``, ``composites``, ``enhancements`` and ``areas`` definitions.
+To use external satpy component configuration it is necessary to define
+``satpy_extra_config_path`` in `external_satpy.yaml <https://github.com/ssec/sift/blob/master/uwsift/etc/SIFT/config/external_satpy.yaml>`_::
+
+    satpy_extra_config_path: [directory path]
+
+or it can be defined via environmet variable ``SATPY_CONFIG_PATH`` as described `here <https://satpy.readthedocs.io/en/stable/config.html#config-path-setting>`_.
```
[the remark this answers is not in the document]
nedelceo: Thanks for your changes, but before a noticed this suggestion I pushed f7287c405c5490d80b1abde391a9b46e1f4f02a4 and (because of it?) I could not commit your suggestion. So I made 180386c1e198c88ba45ee5cc8d12ebac9ad3bc05 to include your version. Sorry for the mess.

uwsift/__init__.py
```diff
@@ -87,10 +89,11 @@ def find_spec(cls, fullname: str, path=None, target=None):
...
 
-satpy_extra_readers_import_path = config.get("satpy_extra_readers_import_path", None)
-if satpy_extra_readers_import_path is not None:
-    sys.path.insert(0, satpy_extra_readers_import_path)
-    os.environ["SATPY_CONFIG_PATH"] = satpy_extra_readers_import_path
+
+# get satpy_config_path from config file
+satpy_config_path_yml = config.get("satpy_extra_config_path", None)
+if satpy_config_path_yml is not None:
+    satpy.config.set(config_path=satpy.config.get("config_path") + [satpy_config_path_yml])
```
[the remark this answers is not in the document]
nedelceo: > we are used to have the python part of the extra readers in the same folder as the yaml part 

Aha, that makes sense now.

uwsift/__init__.py
nedelceo: OK, Satpy alone behaves the same as I wrote above. Since there haven't been any complains, I will resolve this conversation.

nedelceo:
pre-commit.ci autofix

ameraner (approved):
LGTM!

djhoese:
Looks like there is one relevant test failure.

```
 uwsift/tests/test_satpy_import.py:115: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

xdg_config_home = '/home/runner/.cache/SIFT/workspace/temp/pytest-of-runner/pytest-0/test_config_satpy_import_path0/config'
base_config_dir = '/home/runner/.cache/SIFT/workspace/temp/pytest-of-runner/pytest-0/test_config_satpy_import_path0/config/SIFT/config'
overwritten_satpy_import_path = '/home/runner/.cache/SIFT/workspace/temp/pytest-of-runner/pytest-0/test_config_satpy_import_path0/satpy'
satpy_init_path = '/home/runner/.cache/SIFT/workspace/temp/pytest-of-runner/pytest-0/test_config_satpy_import_path0/satpy/__init__.py'

    def check_uwsift_paths(
        xdg_config_home: Union[str, None],
        base_config_dir: str,
        overwritten_satpy_import_path: Union[str, None],
        satpy_init_path: str,
    ):
        if xdg_config_home is None:
            modified_env = {}
        else:
            modified_env = {"XDG_CONFIG_HOME": xdg_config_home}
    
        # Use stderr, because other print statements will confuse the `literal_eval`.
        # If an import is overwritten, then `overwrite_import` will output to stdout.
        command = [
            get_python_path(),
            "-c",
            "import sys, uwsift, satpy, satpy.readers\n"
            "print({'base_config_dir': uwsift.USER_CONFIG_PATHS[0], "
            "'overwritten_satpy_import_path': uwsift.config.get('satpy_import_path', None), "
            "'satpy_init_path': satpy.__file__})",
        ]
        working_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), "..", ".."))
        process = subprocess.run(
            command, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, env=modified_env, cwd=working_dir, check=True
        )
        results = literal_eval(process.stdout.decode("utf-8"))
    
        # Can we configure the config file path with the environment variable `XDG_CONFIG_HOME`?
        assert results["base_config_dir"] == base_config_dir
        # Is the Satpy module path correctly read from the config file `general_settings.yml`?
        assert results["overwritten_satpy_import_path"] == overwritten_satpy_import_path
        # Where is the `__init__.py` from Satpy located?
>       assert results["satpy_init_path"] == satpy_init_path
E       AssertionError: assert '/usr/share/m...y/__init__.py' == '/home/runner...y/__init__.py'
E         - /home/runner/.cache/SIFT/workspace/temp/pytest-of-runner/pytest-0/test_config_satpy_import_path0/satpy/__init__.py
E         + /usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/satpy/__init__.py
```

Any idea if this new output is expected?

nedelceo:
Well I am not able to figure out which change in `__init__.py` is responsible for the AssertationError. Maybe adding `import satpy`?

djhoese:
Due to the way the test is setup it is hard to tell which exact check if failing. The test will change some stuff, run a check, change some stuff, run a check, and so on, but we don't know which set of changed stuff is triggering it since the traceback doesn't seem to say that. Let me try downloading your branch and see if I understand the issue.

djhoese:
Ok I made a commit (but github isn't currently showing it for some reason, we'll wait and see if it shows up today) that seems to fix the tests but I'm not happy about it. You/we were correct that the `import satpy` is messing things up, but that is actually what I wanted to get revealed by these changes. That is, changing `sys.path` to allow the user to customize import paths at import time is extremely fragile and I'd like to avoid this in the future if we can.

I have a meeting now, but I'll try to look at this more today.

djhoese:
@ameraner What is the main use case for `satpy_import_path`? I found the documentation here:

https://sift.readthedocs.io/en/latest/configuration/external_satpy.html#replacing-satpy-by-external-installation

But is this mostly for developers that are running development Satpy with a pre-built SIFT application/bundle?

ameraner:
> But is this mostly for developers that are running development Satpy with a pre-built SIFT application/bundle?

Yes, exactly. This was before my time, but I don't think we requested it - it was implemented as a quick way to try out a dev-Satpy inside SIFT. 

If it's causing problems, I think we could remove this - a developer/expert user most likely runs SIFT via a conda environment, where Satpy can be installed e.g. in pip-editable mode easily...

djhoese:
I think I agree. Sure, yes, there are times when you have a SIFT tarball and you want to test a few hacks to see how it fixes things...but these are all things that can:

1. Be done with modifying the internal `.py` files of a `conda-pack`'d environment now.
2. Should require a unit test for any application type thing and should be done outside the tarball anyway.

Ok. I'll see how hard it is to remove the logic and the documentation.

djhoese:
Ok I removed it and tried to do it all in one commit. The test was updated to basically now test only XDG_CONFIG_HOME which is a decently useful test in my opinion. It is still *a lot* of test code for a simple function though so I'd be open to cleaning it up even further if someone wanted.

djhoese:
Any last issues with this before I merge it (later today)?

ameraner:
LGTM, thanks for your contributions!

arcanerr:
> @ameraner What is the main use case for `satpy_import_path`? I found the documentation here:
> 
> https://sift.readthedocs.io/en/latest/configuration/external_satpy.html#replacing-satpy-by-external-installation
> 
> But is this mostly for developers that are running development Satpy with a pre-built SIFT application/bundle?

Sorry for not contributing to this discussion earlier, but having the option to use a different Satpy version than coming with a bundled SIFT software package was an explicit requirement by EUMETSAT.

I think, the use case was not only for SIFT developers but probably more for external users of SIFT who want to use a newer (or an older) Satpy version which might be better suited for some data to be worked with.

djhoese:
@sjoro Any memory of this requirement mentioned above?

@arcanerr @ameraner In my opinion this shouldn't be allowed. There are too many chances for a version of Satpy that isn't tested with SIFT to break something and not work. I guess the idea is that SIFT will have a slow/delayed release cycle so users should be able to use new readers that have been developed in Satpy since the last release...I don't know. It is too hacky in my opinion.

arcanerr:
I don't see why it shouldn't be allowed to exchange Satpy by configuration even if sometimes things break with a different version if this is clearly communicated (including: "loss of warranty", so to speak). It enables users to do things that they otherwise cannot do. Just my 2 cents.

But if it makes SIFT unstable even if the bundled Satpy is used, I agree, that must not be the case.

djhoese:
Yeah, that's what my "...I don't know" was meant to convey. I like the blanket/generic feature request of "I should be able to use a new version of Satpy with my SIFT installation if I choose to". But the way it is/was implemented was only lucky that it didn't break. Mainly, that it depended on Satpy not being imported before the configuration's import path was read and applied/used. This lead to me trying to `importlib.reload` the Satpy modules and once I did that things just got weird. Doing something like modifying `sys.modules` and using `reload` are not things that will be fun to maintain in the future (especially when it is happening at import time).

Someone using a conda-based installation or development installation of SIFT can still upgrade their version of Satpy freely. The bundled SIFT distribution on the other hand is a different style of environment in my opinion. It is a bundled/guaranteed frozen set of SIFT and its dependencies and I don't think that should be manipulated or be allowed to reach outside of itself (import Python modules outside the bundle). That said, I could see a long term path in the future where SIFT's bundle allows updating of SIFT from within the same distribution. Kind of like any GUI application that has an "Update" button and just asks you to restart to use the new version once it is downloaded. This functionality can and probably will involve updating dependencies.

I'm curious who the original target audience was for the feature as that may skew my feelings about it. If it was me or Andrea or Ray (and since we switched to conda-pack bundles) and we just wanted it for testing, I'd say we hack in a `python -m pip install -e /path/to/my/local/satpy`...assuming pip is available inside conda-pack'd environments. It's a tarball so if we want the original version of Satpy back then we delete the extracted directory and re-extract it. If this feature is for people who don't have much programming experience...then I'm tempted to say they shouldn't be doing this at all. This type of feature may also be a sign that:

1. The release of development bundles ("nightly" builds) was not frequent enough. Since these didn't exist during the EUMETSAT development cycles, this makes sense. This is something every commit to master does from GitHub Actions.
2. The creation of development or conda-based installs was too difficult so people needing to do this weren't comfortable and wanted something that seemed simpler to them. This may be slightly easier now that it used to be, but is still generally an issue.

sjoro:
hi all, a bit of background for this.

basically the EUM requirement for SIFT is/was "I should be able to use a new/another version of Satpy with my SIFT installation if I choose to". the implementation of this is left to the developers and in general, this is not something that should concern any "regular" users.

the reason behind this requirement is that we use SIFT for satellite/instrument commissioning, which is usually rather fast paced period where all the product validation and checks need to be completed. this is a 100% EUM activity during which we can't ecpect and rely on "external" parties, in this case Pytroll/Satpy/SIFT developers, to necessarily react, package and publish new updates to Satpy/SIFT in the timeframe required for commissioning. we have to have a way to bypass official Satpy repo with another version where we can control the environment. later on, once time allows, any beneficial changes to Satpy would be contributed.

sjoro:
@djhoese sorry for not replying, i was on holidays...

the intended users are scientists and/or engineers and we aim for having a central installation so that they do not need to worry about installation details. i would say that a conda-based environment with an option to to development/source install of Satpy is fulfilling the need perfectly.

djhoese:
> is fulfilling the need perfectly

Your first sentence did not make me think this was going to be in your second sentence. I'm glad that that will be good enough. I think with the scripts we have and maybe with more advanced (possibly docker image-based) building processes it wouldn't be hard for someone on the EUMETSAT side to produce the distributed tarball with a development version of Satpy rather quickly if that ends up being easiest for the users you had in mine.

Ok so I think we've decided that the removal of the satpy import path configuration option is OK and we'll have other ways of dealing with it in the future. If it needs to be added back in the future we can talk about that at that time.

djhoese:
Thanks @nedelceo! Merging now.

PR #423, opened 2024-08-29
Fix pre-commit with types-setuptools and limit python to 3.11 due to ecmwflibs

ameraner:
This PR fixes the pre-commit dependencies to move away from the deprecated/yanked `types-pkg-resources` to `types-setuptools`, as suggested in e.g. https://pypi.org/project/types-pkg-resources/ . This was making the pre-commit run fail.

On top, it limits python to 3.11, since there are currently no wheels for `ecmwflibs` for 3.12 and it is preventing the builds to resolve the environments.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -226,7 +226,7 @@ setup(
         "cfgrib",
     ],
     tests_requires=["pytest", "pytest-qt", "pytest-mock"],
-    python_requires=">=3.8",
+    python_requires=">=3.8, <=3.11",  # limiting to 3.11 until ecmwflibs is not available for 3.12
     extras_require=extras_require,
     packages=find_packages(),
     entry_points={
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -128,7 +128,6 @@ jobs:
           --no-deps --upgrade \
           git+https://github.com/dask/dask \
           git+https://github.com/dask/distributed \
-          git+https://github.com/zarr-developers/zarr \
           git+https://github.com/Unidata/cftime \
           git+https://github.com/mapbox/rasterio \
           git+https://github.com/pydata/bottleneck \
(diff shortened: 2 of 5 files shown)


ameraner:
Not sure why the experimental test env is failing with 
```
../../../miniconda3/envs/test-environment/lib/python3.10/site-packages/zarr/codecs/crc32c_.py:7: in <module>
    from crc32c import crc32c
E   ModuleNotFoundError: No module named 'crc32c'
```
if the other ones are ok? The experimental build is using the same py3.10 and ubuntu-latest as the normal test run...

djhoese:
So I removed zarr from the unstable dependencies. I did the same in pytroll's projects because it ends up needing a lot of unique dependencies that no other packages need. With the way unstable deps are installed we'd have to list each one of zarr's dependencies which is just annoying. Now the unstable CI is failing "as expected" because numba doesn't support new numpy.

setup.py
```diff
@@ -226,7 +226,7 @@ def run(self):
         "cfgrib",
     ],
     tests_requires=["pytest", "pytest-qt", "pytest-mock"],
-    python_requires=">=3.8",
+    python_requires=">=3.8, <=3.11",  # limiting to 3.11 until ecmwflibs is not available for 3.12
```
djhoese, line 229: Is the extra space allowed? `>=3.8,<=3.11`?

ameraner: Should be yes, this https://packaging.python.org/en/latest/specifications/version-specifiers/#id5 says 

> Whitespace between a conditional operator and the following version identifier is optional, as is the whitespace around the commas.

ameraner:
Great, thanks for the fix!

ameraner:
Issue related to the ecmwflibs py3.12 wheels: https://github.com/ecmwf/ecmwflibs/issues/33

PR #438, opened 2025-02-25
Add python 3.12 tests to CI

ameraner:
Add test runs for python 3.12 now that it is a supported version.


ameraner:
See also https://github.com/ecmwf/ecmwflibs/issues/33 for the move from pip to conda

ameraner:
Hmm, looks like putting eccodes as pip dependency breaks the deploy since eccodes is likely a dependency of some other conda package now, so it gets installed by conda first and then overridden by pip, which breaks python-eccodes... But keeping everything in conda was also no making the build happy as it could not find the package (although tests were ok(?)). Will have to try again.

PR #439, opened 2025-02-25
Reactivate (and possibly fix) export image and rgb config tests

ameraner:
As described in https://github.com/ssec/sift/pull/437, this PR reverts the deactivation of the image export and rgb tests, in order to be able to monitor more easily if the are passing again and/or to work on a fix.


ameraner:
`test_rgb_config.py` was reactivated by https://github.com/ssec/sift/pull/448 after it started working again without intervention.

ameraner:
`test_export_image.py` could be reactivated as the Segfaults disappeared, but had to limit imageio and pyav due to https://github.com/imageio/imageio/commit/f017778434614b1f03c5df75f0e5f0b5560f86a5#diff-495dd3bb6bee17e7f0aec89db20d74e14824891249ceba5ab66950287801bb2d causing https://github.com/imageio/imageio/issues/1120 (then triggering https://github.com/imageio/imageio/issues/1111)

PR #441, opened 2025-02-26
Define texture internalformat to fix colormap discretisation

AnaVukasinovic:
The bug occurred because the "internalFormat" field of the data texture was left as None. The bug appeared in the SIMPLE_GEOLOCATED and PIXEL_MATRIX ImageDisplayMode. It was fixed by overriding the Image class, where the _init_texture method was overridden to set the "internalFormat" parameter to "r32f"


uwsift/view/scene_graph.py
```diff
@@ -87,6 +87,13 @@
 DEFAULT_TEXTURE_SHAPE = (4, 16)
 
 
+class CustomImage(Image):
+    """Custom image override to fix issue #404."""
+
+    def _init_texture(self, data, texture_format, **texture_kwargs):
+        return super()._init_texture(data, texture_format, internalformat="r32f")
```
ameraner, line 94: maybe it's more futureproof if we still pass on eventual texture_kwargs as in below, if you agree?
```suggestion
        return super()._init_texture(data, texture_format, internalformat="r32f", **texture_kwargs)
```

ameraner (commented):
Just a suggestion inline

PR #443, opened 2025-03-07
Initial commit for bug 392

AnaVukasinovic:
The problem was that numerical values in the table in the Open File Wizard window were treated as strings during sorting. The bug is resolved by overriding the QTableWidgetItem class, specifically the __lt__(self, other) method. This way, numerical values are correctly compared and sorted, while simultaneously being displayed in the appropriate format in the table.

### DIFF
--- a/uwsift/view/open_file_wizard.py
+++ b/uwsift/view/open_file_wizard.py
@@ -74,6 +74,28 @@ RESAMPLING_METHODS = {
 }
 
 
+class NumericTableWidgetItem(QtWidgets.QTableWidgetItem):
+    """Custom QTableWidgetItem class to override the __lt__ method
+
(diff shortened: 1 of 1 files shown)


RudiLauster:
Reviewed and tested, all fine from my side!
After Ana @AnaVukasinovic has removed the DRAFT state, this can go to Andrea @ameraner 👍

RudiLauster:
> Reviewed and tested, all fine from my side! After Ana @AnaVukasinovic has removed the DRAFT state, this can go to Andrea @ameraner 👍

I forgot to check why the coverage analysis has failed. So, hold on, investigating ...

ameraner:
LGTM, works great, thanks!

ameraner (approved):
LGTM, works great, thanks!

PR #444, opened 2025-03-13
Implement resources release and scene deletion to solve segfault crashes when reusing files

### LINKED ISSUE #415: SIFT tends to crash when loading datasets from the same FCI repeat cycle in separate Open File Wizard calls
**Problem**
To reproduce the issue:
- In the Open File Wizard, open the 40 chunks for an FCI repeat cycle
- select one or more datasets to open
- after they opened, reopen the Open File Wizard and select one or more datasets from the same repeat cycle
After this, and during the Scene opening and loading operations, SIFT has been observed to crash and disappar with a Segfault. However, this does not happen every time consistently. 
This has been observed on a Linux system.

**Analysis**
The suspect is that it is an issue with multiple accesses to open NetCDF files. Since the Scene gets reinitialised at every Open File Wizard call anyway, possibly a better cleanup/closing/deleting/garbage-collecting of Scene/Filehandlers might fix the issue.

**Workaround**
Make sure to load all needed datasets from a single repeat cycle in one Open File Wizard call.

Possibly related to https://github.com/pytroll/satpy/issues/2092

roche-emmanuel:
Hi everyone!

@ameraner , @strandgren ,

In this pull request, I'm providing the changes I used on my side to resolve the issue #415: 

In fact I could very easily reproduce this issue myself (as I'm mostly working from Windows), so, initially, I got crashes about 50% of the times I was trying to re-open the same input files to load additional datasets. With these fixes, I don't seem to be able those crashes anymore.

There were in fact a couple of locations in SIFT where a satpy scene is created: first in the OpenFileWizard, where the scene is created to collect the list of available datasets for the following page display, and then in the SatpyImporter class, where another scene is created to effectively load the data.

- In the case of the open file wizard, the loaded scenes would never get garbage collected since there were referenced in the wizard object, and that object is itself referenced in the Main class of the app. So I have now added a "reset_state()" method on this wizard, to ensure that, when done with the file loading process, we effectively unload the satpy scene and garbage collect it to release the associated file resources correctly.

- In the case of the SatpyImporter a scene is referenced in the class (self.scn) and then used in the simple_workspace to load actual data layers when we call `gen = truck.begin_import_products(prod)` (truck is the importer here). So the solution I'm suggesting here is to add another virtual method in the base aImporter class called `release_resources()`, and reimplemented to unload the satpy scene correctly in the SatpyImporter. Then we simply call  `truck.release_resources()` at the end of the `import_product_content()` in the workspace.

=> So please, let me know if you have any feedback on this or any question/point you think we should discuss further or change before this can be merged in.

Thanks a lot in advance for your time reviewing this 🙏😊!

Closes #415

### DIFF
--- a/uwsift/__main__.py
+++ b/uwsift/__main__.py
@@ -346,6 +346,11 @@ class Main(QtWidgets.QMainWindow):
         """
         self.didFinishLoading.emit(uuid_list)
 
+        if self._wizard_dialog is not None:
+            # Issue #415: reset the state of the wizard here to release
+            # the satpy scenes:
+            self._wizard_dialog.reset_state()
+
         if not uuid_list:
             raise ValueError("no UUIDs provided by background open" " in _bgnd_open_paths_finish")
         if not isok:
@@ -1143,6 +1148,10 @@ class Main(QtWidgets.QMainWindow):
             self.open_paths(self._wizard_dialog.files_to_load, **importer_kwargs)
         else:
             LOG.debug("Wizard closed, nothing to load")
+            # Issue #415: We still need to reset the state of the wizard here
+            # as a satpy scene might have been loaded already even if the user cancels
+            # the opening process.
+            self._wizard_dialog.reset_state()
 
     def _reload_config(self):
         config.refresh()
--- a/uwsift/view/open_file_wizard.py
+++ b/uwsift/view/open_file_wizard.py
@@ -14,6 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with SIFT.  If not, see <http://www.gnu.org/licenses/>.
+import gc
 import logging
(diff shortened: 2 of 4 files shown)


strandgren:
Hi @roche-emmanuel, I just tested your branch and I cannot reproduce the issue with the segmentation fault 🙂  When switching to the master branch it keeps happening in about 50% of the cases.  So to me it seems like your changes fixes the issue 👏 However, we should also wait for a final review by @ameraner next week

ameraner:
Thank you for this nice work, indeed I can also confirm that it seems to work well and solves the issue!

ameraner (approved):
LGTM, and it works great, thanks! This is a really nice improvement in the usability and stability!

Issue #333, opened 2022-04-28
Export Image with Colorbar : colorbar doesn't flip when enhancement flips


ScottLindstrom:
I loaded a level 2 product -- Cloud Particle Size Distribution -- and applied a color bar to it.  Exported it, with a color bar.  All a-okay.
I flipped the color bar, instead of 0 to 100, made it 100 to 0.  Exported the image, with a color bar.  The color bar was the original colorbar, it didn't flip.
Same behavior happens with L1b imagery.

djhoese:
This seems to still be a bug in the EUM version of SIFT.

CC @sjoro @ameraner @strandgren

PR #445, opened 2025-04-01
Issue #333: Fixing color limit inversion support for image export

roche-emmanuel:
Hi @ameraner , @strandgren , @sjoro ,

This PR provides the required fixes for issue #333: 

To support the "color bar inversion" while generating the exported image we simply need to always map our color limit min/max values appropriately (ie. irrespectively of the name "vmax/vmin" that we use for those values): this part of the fix is included in the export_image.py file.

Now, on a similar not, in the layer_details.py conversions between QT physical units and dataset value units were not always properly taking into account that what we call "valid_max"/"valid_min" could also be inverted. So I introduced a few changes in that file too. 

=> Overall, the image export feature now seems to work as expected for me with those changes. But of course, please let me know if you think there is anything we need to discuss/change before the PR can be accepted. Thanks 🙏!

### DIFF
--- a/uwsift/view/export_image.py
+++ b/uwsift/view/export_image.py
@@ -257,7 +257,11 @@ class ExportImageHelper(QtCore.QObject):
 
         cmap = mpl.colors.ListedColormap(colors)
         vmin, vmax = self.model.get_dataset_presentation_by_uuid(u).climits
-        norm = mpl.colors.Normalize(vmin=vmin, vmax=vmax)
+        # Issue #333: need to take into consideration that the vmin/vmax order might be
+        # inverted here:
+        mini = min(vmin, vmax)
+        maxi = max(vmin, vmax)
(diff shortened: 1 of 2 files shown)


RudiLauster:
Hello Andrea @ameraner, hello Manu @roche-emmanuel,

I also took a look at this PR using a FCI L1 product and could reproduce that the bug is gone! 👍🙂👍

Besides:
Andrea, I found two issues that are not really related to this bug and I wonder if they have already been detected: I loaded two layers and found out that only the colour bar of one of the loaded layers is exported into the image.
No matter what layer I select, always only this one colour bar is in the image.
Also I selected a radiance map and the colour bar scale is crowded and unreadable in the image exported.
Both of those issues are already on the master, so again, have nothing to do with this PR.

ameraner (approved):
LGTM and works well, thanks!

PR #446, opened 2025-04-10
Inverse toggle buttons for "Fit to data"

### LINKED ISSUE #368: Add an "inverse" toggle to "Fit to data" (at least for terrestrial IR channels)
When applying the "linear" colormap to terrestrial IR channels, we can either set minimum < maximum for dark clouds, or maximum < minimum for bright clouds. The latter is more conventional. When the user chooses "Fit to data", the lowest BT becomes black and the highest BT becomes white (minimum < maximum).

It would be nice to have a little checkbox "invert", that would invert the fitting process, such that the lowest BT becomes white and the highest BT becomes black. This checkbox could be located on the same line where the text "Fit to data:" and the buttons "Current time" and "All times" are located.

Another option for the same effect would be a general option to invert the colormap.

RudiLauster:
Created feature according to what was discussed in https://github.com/ssec/sift/issues/368

Also introduced button states to actually see if the inverted (or non-inverted) Fit-to-data is currently active. The button is available again when a different Fit is selected or when the user changes the ranges manually.

Also due to adding this functionality, the Fit-to-data button layout had to be improved for readability (and aesthetics 😉)

Closes #368


RudiLauster:
Hello Andrea @ameraner ,

Here is another PR from Ana, I have already reviewed and tested it on my side. So, over to you 😉

Just saw that the linting checks complain about some issue that is not related to this PR: https://github.com/ssec/sift/actions/runs/14377944883/job/40314732342?pr=446
This issue must be there since long time, according to Git blame. Maybe someone has tightened the rules? 😉
But I think this might be easily fixed. The error message says "... name is never assigned in scope": So I assume that if the 
...
# singleton instance
_MDB = None
...
would be simply moved to the class Metadatabase itself (where in a way it semantically belongs to), the problem might be gone.

Anyway, you have the mic now 😁

RudiLauster:
Hello Andrea @ameraner,
so, Coveralls is happy again - over to you 🙂

ameraner (approved):
LGTM, thanks!

Issue #406, opened 2024-01-12
Allow more than two decimals for deciding the edges


gerritholl:
When deciding the edges for the visualisation, sift allows only two decimals.  However, when data are tightly spaced, a more precise specification would be desirable.

For example, here is the 0.9/0.8 ratio for a specific region, where I would like to define edges between 0.0095 and 0.0115.  The closest I can do is between 0.00 and 0.02.  It's also too much discretised, but that is the same problem as described in #404.

![grafik](https://github.com/ssec/sift/assets/500246/0a7b7bc4-9109-483e-8de3-25acae9ea202)

ameraner:
A more extreme example is given by @strandgren in EUM Gitlab issue 172:

![image](https://github.com/ssec/sift/assets/49722768/382e664e-0f0b-44ce-b8f5-043394c85cb7)

PR #450, opened 2025-05-08
Issue 369: Adaptive increases for ranges arrows

RudiLauster:
Implemented as discussed:
The increment / decrement via the arrow buttons or the mouse wheel will be done on the last digit displayed in the edit field.
That means if you add digits to the value e.g. by adding "0" to the end, those will be the significant ones to be incremented / decremented.
So, implicitly the issue #406 "Allow more than two decimals for deciding the edges" has been resolved with this PR, too.
Hence, it is possible to use Ctrl and/or Shit as modifiers to increment or decrement byfactor 10, 100 or 1000. See the tool tips when hovering over the spinbox.


RudiLauster:
Hello Andrea @ameraner ,
The new class QAdaptiveDoubleSpinBox in this PR yet lacks a new unit test, but maybe you can already check if this is what you would expect?
The PR is based on PR https://github.com/ssec/sift/pull/448, so actually only the last commit determines the significant changes for this PR.

RudiLauster:
Hello Andrea @ameraner ,

I have added a unit test and so this PR is now also ready for review 👍

RudiLauster:
After trying to bring this PR together with Ana's Fix for 406, some changes for rounding correctly were necessary.

ameraner:
Merged as part of https://github.com/ssec/sift/pull/452

Issue #414, opened 2024-05-07
Add full-res image export functionality


ameraner:
**Problem**
The current SIFT export image functionality basically takes a simple screenshot of what is currently shown in the image area, at the desktop resolution, and saves it to file.

**Feature Request**
A user request at the last Convection Working Group (9th CWG Recommendation 2), asks for the functionality of exporting what is in the image area (up to) the native full resolution of the data. 

**Additional Note**
If possible, in addition to simple image formats, an export in GeoTiff including the projection information shall be selectable.

ameraner:
I assume, in order of priority for implementation:
- the full-res image of the currently shown/active layers, with the currently respective selected and tuned colorbars
- on top of that the overlays (grid, coastlines), if active in the Layer Manager

I do not think that the polygons/probes would be necessary for the full-res output variant.

roche-emmanuel:
Feeewww... it seems the geotiff projection export is finally starting to work:

Here is an area I just exported to geotiff:

![Image](https://github.com/user-attachments/assets/a73a3fab-0dc7-4ddc-a26c-436f9a62fdbb)

And then here is the exported geotiff image re-loaded from a jupyter script to display it with cartopy coastlines on the full earth disk:

![Image](https://github.com/user-attachments/assets/880ca859-9e16-4c09-a0e3-ba0dc32e4247)

PR #451, opened 2025-05-09
Issue #414 optimal resolution image export

roche-emmanuel:
Hello @ameraner,

Here is the merge request concerning the SIFT issue #414:

---

### 1. Output Resolution Selection

I've now added support for selecting the output resolution when exporting an image, with the following options:

* **"Default"**: Uses the current on-screen canvas resolution (this is the original behavior).
* **"Optimal"**: Determines the maximum resolution available among all currently loaded datasets, then relates that to the visible portion of the scene to estimate how many pixels are being shown from the highest resolution source (both width and height). This estimated resolution is then used for the export.
* **"Custom"**: Allows the user to specify a custom resolution, while maintaining the current aspect ratio of the canvas display.

---

### 2. GeoTIFF Export Support

I've also been working on GeoTIFF export support, which—honestly—is where I’ve spent most of my time so far 😅.

To support this, I added an additional image type option: `"*.tif"`, which the user can select when exporting an image.

When this is chosen, the export process will additionally:

* Visit the scene to find the first image node that provides a `Proj4Transform`, and retrieve its Proj4 description string.
* Use `rasterio` to write the GeoTIFF using the projection information.

The basic functionality is *kind of* working already (i.e., we’re generating GeoTIFF files), but things start to get tricky when it comes to specifying the transformation details for the projection embedded in the file.

For example, suppose the projection is:

```
'+ellps=WGS84 +h=35786400 +lon_0=0 +no_defs +proj=geos +type=crs +units=m +x_0=0 +y_0=0'
```

(This is what I’m currently reading from the default FCI test input files.)

This projection expects lat/lon coordinates as inputs. However, a GeoTIFF file typically only supports affine transformations, which assume linear spacing between pixels. That’s a problem, because latitude and longitude don’t progress linearly across the image pixels.

Another issue is that the corners of the exported image could lie outside the valid bounds of the projection. In that case, we might be able to extrapolate the valid lat/lon range and still expect the projection to work correctly—for example, if a pixel's longitude is 123.4°, the forward projection might return NaN (I think?).

---

### Questions

So, the only way I currently see to make this projection export useful would be to include the **lat/lon arrays** corresponding to the exported image area within the GeoTIFF file itself.

Is that what you had in mind? Or were you thinking of something else?

Also, this only applies in this particular case (i.e., when the projection expects lat/lon as input). Other SIFT inputs might involve different projection types, so supporting GeoTIFF export generically could become even more complex.

So I was wondering if you could provide some additional indications on how you think we should proceed on this point (ie. properly exporting the projection transformation for any image position, and any input projection) ?

(**Side note:** If there’s anything you'd like to discuss or change regarding **Point 1** as well, feel free to let me know 😊)


djhoese:
Could you describe to me in more detail what projection is used? From what I would expect (as a user), if I export the current Canvas/Map to a geotiff then I would expect it to use the projection of the current view, not of the data, take a "screenshot" of the canvas, and save the geotiff. In this case I only see issues if the user is zoomed out a lot (and we're getting a lot of space pixels) and the extents are invalid as you pointed out, but the projection shouldn't be an issue unless I'm missing something. How does my assumption differ from how it actually works here?

> For example, suppose the projection is:
>
>> '+ellps=WGS84 +h=35786400 +lon_0=0 +no_defs +proj=geos +type=crs +units=m +x_0=0 +y_0=0'
>
> (This is what I’m currently reading from the default FCI test input files.)
>
> This projection expects lat/lon coordinates as inputs. However, a GeoTIFF file typically only supports affine transformations, which assume linear spacing between pixels. That’s a problem, because latitude and longitude don’t progress linearly across the image pixels.

I'm not sure I understand what you mean here. The "geos" projection's coordinate system is in meters typically (sometimes radians, but let's ignore that case). With `pyproj.Transformer.from_crs` (or similar) you can transform coordinates from a lon/lat projection to the geostationary projection, but the lon/lat projection isn't a requirement. For example, you could transform coordinates from mercator x/y meters to geostationary x/y meters. The legacy pyproj interface `Proj` does do lon/lat <-> some projection, but this interface should probably not be relied upon in the long term as I believe it is considered deprecated.

roche-emmanuel:
Hello @ameraner and @djhoese ,

First thanks a lot for your feedback on this and sorry for the delayed answer ;-) (I was really busy with other stuff lately).

Actually, the perspective you shared above Dave might be the key to unlock this situation: I'm not sure why in the first place, but I was thinking I should try to retrieve the original projection of the source data so this is where I was facing this lat/lon input consideration (plus we have to add on top of that a good share of confusion from my side as I've always been pretty confused with proj4 projections 😅). 

But, if instead we focus on "using the projection of the current view, not of the data" as you suggest then we should (probably) get a linear grid of coordinates, which I could use to specify the projection affine transform in the geotiff file🤔(?) (Note: this also resolve the other question I had in mind with "multiple possible input projections": we just use the geostationary grid every time in the output file)

=> So, I'm going to revisit this now and will let you know if I can make any progress (note: I'm on vacation on thursday/friday, so I don't think I can resolve this this week, and will continue on it next week).

(Also, I'm now realizing I should be able to test the validity of the output simply trying to load the geotiff image back into a simple python script to try to plot them on the full earth geos projection for instance, so I will do that).

roche-emmanuel:
Hi @ameraner , @djhoese ,

I finally got a chance to look again into this geotiff projection issue and I think I now have an acceptable solution implemented in the MR.

As mentioned above, I could use the exported tif images to try to reload them somewhere else checking that the projection infos would match the expectations: so this is what I've done now and I posted a few images directly in the issue #414 discussion.

And as already discussed, the main concern then was on "how to get proper image bounds for views where we don't have any valid data at the borders of the current view. So to handle this, I use the (presumably acceptable) assumption that the user will at least have some valid earth pixels around the center area of the image [otherwise we fallback to exporting the tif with no projection data]. From that center area we collect the lat/lon coords for 4 points, transform them with the geos transform, and then extrapolate those values to match the full size of the image (potentially producing values that are invalid at the border of the images here [but still finite, so usable for linear interpolation later on]).

Then when you read back the image data somewhere, depending on where you use it you might need to extract the valid part of the extrapolated bounds of the image to display it correctly (this is the case with the test script I used myself to display the image with matplotlib and cartopy).

Not sure if this is exactly what you had in mind, but that's the best solution I could find so far and at least this seems to work fine ;-).

Anyway, I think this PR should now be ready for a review from your side, so of course, let me know if you think there is anything that should be changed or you have any question on the modifications I introduced here.

Thanks for your time on this 😉🙏!

roche-emmanuel:
Hello @ameraner 👋!

So, as requested above, I have now also added the support for the ".tiff" extension, and untroduced some additional unit tests on the tif writing code in export_image (and now the code coverage is slightly increasing instead of decreasing).

Now as far as I can see the only failing test on the repo is the build with [(3.11, ubuntu-latest, true) (pull_request)](https://github.com/ssec/sift/actions/runs/15631346735/job/44036224769?pr=451) which seems to be due to an incorrect numpy version, so I don't think it can be related to the changes introduced in this PR?.

=> Anyway, let me know if you need me to change anything else on this 😊.

ameraner:
Hi @roche-emmanuel , thanks a lot for this! LGTM. 
The failing test is the experimental one that is "allowed" to fail (see the "true" in the action name) as it tests the latest libraries experimentally.

PR #452, opened 2025-05-16
Gmv issue 406

AnaVukasinovic:
This pull request contains part of the implementation for bug 406, which relates to the additional functionality of adjusting the limits on sliders by clicking the Reassign button. Specifically, the values currently selected on the sliders, when clicking the Reassign button, become the min and max values, i.e., the slider boundaries. 

This pull request relies on Pull Request #448, which enables variable min and max values, as well as Pull Request #450, which approves a larger number of decimals in the spinbox values.


ameraner:
Hi  @AnaVukasinovic , @RudiLauster , 

thanks a lot for the PRs! I understand this PR superseeds/includes https://github.com/ssec/sift/pull/450 and https://github.com/ssec/sift/pull/448, so those can technically be closed if we merge this one, right?

The new sliders and buttons behave generally very nicely, thanks for all the work on it!

Some points here:
- when I scroll over the maximum slider, somehow the focus goes to the gamma field (the number is highlighted in blue, and if I type a number, it is entered in the gamma). Can you reproduce this? Any idea?
image

- when I click on a Fit to data button, the ranges now update to _a lot_ of digits. Is there any way to limit this, e.g. by keeping the previous number of digits (or any other idea)? Like this it's a bit annoying since the increment is then very small (the float precision likely)
image

- the panel is becoming a bit crowded with labelled buttons and a bit confusing... I am wondering, is it easy to move the "Reassign" and "Reset" buttons to the right of the numbers, and use icons instead? A bad drawing below just as example.. what do you think?

image

RudiLauster:
Hello Andrea @ameraner !

> thanks a lot for the PRs! I understand this PR superseeds/includes #450 and #448, so those can technically be closed if we merge this one, right?
Yes, they are actually contained in this one 🙂

 
> The new sliders and buttons behave generally very nicely, thanks for all the work on it!
Great, thank you! 😀

> Some points here:
> 
>     * when I scroll over the maximum slider, while the value for the slider changes, somehow the focus goes to the gamma field (the number is highlighted in blue, and indeed if I type a number, it is entered in the gamma). Can you reproduce this? Any idea?
Yes, I can reproduce this. I am afraid that this is a side-effect of the workaround that inhibits the "double increment/decrement when extending range". But this is just a suspicion, so need to look at that and check if this can be solved quickly ...

>     * when I click on a Fit to data button, the ranges now update to _a lot_ of digits. This is annoying since then it also defines the resolution on the increase/decrease steps. Is there any way to limit this, e.g. by keeping the previous number of digits (or any other idea)? Like this it's a bit annoying since the increment is then very small (the float precision likely)
Jupp - we have overseen that ... just solved that with the last commit 👍Thank you for seeing this!

>     * the panel is becoming a bit crowded with labelled buttons and a bit confusing... I am wondering, is it easy to move the "Reassign" and "Reset" buttons to the right of the numbers, and use icons instead? A bad drawing below just as example.. what do you think?
You're right, it has become to be a bit confusing...
I am a fan of icons, though it can be quite difficult to design them if the meaning deviates from "standard semantics". But "Reset and "Fit" are definitely known semantics.
I guess you thought of icons in order to save space. And looking at that I had some other idea to save space and keep the textual buttons. The Gamma takes quite some space and putting it to the right makes it have a reasonable size. Then the Reassign and Reset buttons would be below in a logical context of the sliders and spin boxes. What do you think?
image

ameraner:
Thanks for checking!

> I guess you thought of icons in order to save space. And looking at that I had some other idea to save space and keep the textual buttons. The Gamma takes quite some space and putting it to the right makes it have a reasonable size. Then the Reassign and Reset buttons would be below in a logical context of the sliders and spin boxes. What do you think?

Looks good I think! The gamma is rarely used so I'm ok with giving it less prominence. With this, we could also remove the separator line above `Fit to data:` and save a little more space + make it clear that the  `Fit to data:` buttons belong to the colormap items as well.

AnaVukasinovic:
Hi Andrea @ameraner ! 
Thank you for your feedback!

In my latest commit, I included a frontend fix in the layer details, and it now looks like this: 

layer_details_frontend

I also removed the bug related to shifting focus to the gamma spinbox. 

Please let me know if you like the frontend in the layer details now, and if the bug with transferring focus to the gamma spinbox has also disappeared on your side.

RudiLauster:
Just fixing some minor 1px layout issue ... please hold on, will commit in a bit 🙂

RudiLauster:
Hi, Andrea @ameraner, so I guess we are ready for a further review from your side 👍🙂

Issue #416, opened 2024-05-07
Link selecting of layer in layer manager with driving layer


ameraner:
**Problem**
During user testing, it was noticed that the current setup and combination of Layer Manager and Driving Layer box is not very intuitive and user friendly; In fact, these are currently completely unlinked, so that a user needs to actively manipulate both to see the wished information. 
However, a novel user intuitively expects, after having clicked on a layer in the Layer Manager, to see the according time steps also in the Timeline Manager. This does not happen per default, since the according item needs to be additionally selected in the Driving Layer box.

**Proposal**
To improve the intuitiveness of this, I would suggest to link the action of selecting a Layer in the Layer Manager with the automatic selection of the according item in the Driving Layer box. 

To maintain the current flexibility, the other way around should not apply; although we could consider going further with this and get rid of the driving layer manager alltogether and command the timeline via the Layer Manager, for simplicty. I cannot think of a drawback of this at this time...

PR #453, opened 2025-06-04
New feature 416

AnaVukasinovic:
This pull request refers to the implementation of a new feature for linking the Layer Manager and the Driving Layer box  https://github.com/ssec/sift/issues/416 :

1. They are initially connected (the same layer is shown in both by default)
2. When a layer is selected in the Layer Manager, that same layer becomes selected in the Driving Layer box. 
3. This functionality also works when adding new layers to the Layer Manager, as well as when adding Algebraic and Composite layers (once they become dynamic, of course).

The other way around is not applied, as stated in the feature request itself, in order to preserve flexibility. That is, when a layer is selected in the Driving Layer box, it is not automatically selected in the Layer Manager.


ameraner:
Looks and works great, thanks!
See new "Link to Selected Layer" checkbox below the driving layer dropdown:
image

PR #455, opened 2025-07-04
Reimplement equalizer view

### LINKED ISSUE #361: Revive "equalizer view"
I thought I had an issue for this already, but can't find it so here we go...

### Removal and old behavior

During EUMETSAT development the layer list was rewritten and the "equalizer view" of each layer was removed due to the complexity and user experience issues that were run into. The equalizer view is/was a visual indication in the layer list to show the value of each layer where the point probe was placed. This was indicated used a blue-ish/purple-ish bar to color the layer list row for each layer. The size/width of the bar was determined by the probed value in the range of the color limits of the layer.

### User Experience Issues

The user experience issue with this is that you could have different layers with different color limits or with similar units on completely different scales. A visualization like this could show two layers as being the "same" but with completely different values being probed. For example, a temperature difference field and a top-of-atmosphere BT layer. Or a sea surface temperature and a TOA BT layer. They may both be showing Kelvin or Celsius but could be on different scales and their "bars" appear to show them at the same value.

This wasn't exactly a problem for old SIFT because UW/CIMSS/SSEC training focused on the Level 1B products and took the "lazy" approach that all BT bands had the same color limits and Visible bands had the same color limits. So two bars at the same position were probably showing the same temperature on the same scale. Worst case you were looking at a visible albedo (% maybe?) value and a temperature, but with these being the two primary types of products we looked at this was easy to distinguish.

SIFT now supports almost every available Satpy reader including Level 2 products, builtin Satpy composites, etc. This can easily get confusing in a simplified visualization like the equalizer view was trying to do.

### Revival

The equalizer view was used heavily by trainers using SIFT and needs to have some equivalent replacement. That is, you need to be able to place a point probe and at a glance see "this is how these layers 'respond'".

The easiest but probably not the best solution would be a new widget (similar to the region probe graphs) that looks just like the old equalizer layer list. All the UX issues, but nothing new for past users.

Another option during a brainstorming session with @rayg-ssec was to maybe display colorbars with some kind of indicator at where that probed value is in the over colors. This doesn't exactly solve the differing color limits/scales issue, but it could still be useful.

We could group layers into types and have individual equalizer bars for each product in that group. The min/max of the plots would maybe be the min/max of all the color limits for the layers in that group. If the grouping is done well then this should "just work" and be pretty obvious. Especially if there are titles for each of the groups to indicate what we're looking at. I can see training language being something like "We can see in the TOA Brightness Temperature bands that many of them have a high value, but the TOA Reflectance bands...". I have no idea if this is actually something that be encountered, but at least the grouping avoids the major issue of the mixing of unlike layers/scales.

RudiLauster:
- Implemented the equalizer bar based on range values (if available) resp. colour limits.
- Optimised the coloumn "size management" - a double click on the column border will reset to a optimal size.
- Items not readable are being provided with tool tips.
- Moved the Statistics Dock to the Layer Details as a dedicated Tab.
- Added unit tests.

closes #361

### DIFF
--- a/uwsift/__main__.py
+++ b/uwsift/__main__.py
@@ -831,6 +831,8 @@ class Main(QtWidgets.QMainWindow):
 
         self.layer_model.init_system_layers()
 
+        self.ui.treeView.resizeColumnsToContents()
+
     def _init_algebraic_pane(self):
         self.algebraic_config_pane = AlgebraicLayerConfigPane(
             self.ui, self.ui.algebraicScrollAreaWidget, self.layer_model
@@ -938,6 +940,7 @@ class Main(QtWidgets.QMainWindow):
         return engine.rootContext()
 
     def _init_arrange_panes(self):
+        self.tabifyDockWidget(self.ui.layerDetailsPaneDockWidget, self.ui.datasetStatisticsPaneDockWidget)
         self.tabifyDockWidget(self.ui.layerDetailsPaneDockWidget, self.ui.rgbConfigPaneDockWidget)
         self.tabifyDockWidget(self.ui.layerDetailsPaneDockWidget, self.ui.algebraicConfigPaneDockWidget)
 
@@ -945,11 +948,9 @@ class Main(QtWidgets.QMainWindow):
         self.ui.layerDetailsPaneDockWidget.raise_()
 
         self.ui.datasetStatisticsPaneDockWidget.show()
-        self.ui.datasetStatisticsPaneDockWidget.raise_()
-
         self.ui.areaProbePaneDockWidget.hide()
-        self.ui.algebraicConfigPaneDockWidget.hide()
         self.ui.rgbConfigPaneDockWidget.hide()
+        self.ui.algebraicConfigPaneDockWidget.hide()
 
         # refer to objectName'd entities as self.ui.objectName
         self.setAcceptDrops(True)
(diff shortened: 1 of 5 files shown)


RudiLauster:
Hello Andrea @ameraner, hello Johan @strandgren,

so, basically the Feature is implemented, but I have put it to draft since I want to find out if there are any issues that the CI will detect.

So, now the Equalizer View looks like this: 
image
image

Do you like the look & feel? If you already have anything you like to be changed, just tell me. Now I have forgotten to go through my notes in order to double check if I have  implemented all we talked about.

There are some config items that are internal and that avoid magic numbers in the code. But I have put them local to the class, is that OK? E.g. COL_WIDTH_MAX, SLIDER_MIN_VAL or SLIDER_MAX_VAL

Also, I grab the ranges as they come from the model. What we talked about, to have them from the default_colormaps.yaml I didn't touch. This seems to be not as implemented as I expexted, e.g. the entry "range" in here has been commented:

default_colormaps:
  toa_bidirectional_reflectance:
    colormap: Square Root (Vis Default)
#    range: [-1.2, 120]

RudiLauster:
Uhh .. well it does not look like that on Windows ... let me check if Windows is somewhat tolerant to a bogous initialisation ...

ameraner (approved):
Thanks for the fix, looks and works great!

PR #466, opened 2025-09-23
Split datasets and composites in Open File Wizard

### LINKED ISSUE #412: Split datasets and composites in Open File Wizard
Now that SIFT can open single datasets as well as composites, the list of products is quite crowded and a bit confusing, since the two types of products are mixed together. I would suggest to split datasets and composites, e.g. using a radio-button or toggle-switch, to switch between the list of datasets and list of composites. Selections from both lists shall be kept.

The current list of all available loadable datasets is retrieved here: https://github.com/ssec/sift/blob/ac7d4315cdf0d34a6900bb618328b087d276f312/uwsift/util/common.py#L107

The list of available composites can be retrieved with `scn.available_composite_ids()`, and the available datasets with `scn.available_dataset_ids(composites=False)`

image

Ported from EUM Gitlab issue 166

RudiLauster:
The function create_scenes now returns all, composite-only and single-only datasets


ameraner (changes requested):
Nice, thanks for this! As commented inline, we need to switch singles and composites.
Also:
- the icons do not show on my linux, for some reason: 
image
- now you're using a QComboBox, but what do you think about a QRadioButton or maybe even better a QPushButton instead, with three buttons and only one can be active? I think it would be a bit more intuitive, prominent in the UI and less clicky-scrolly to use...

RudiLauster:
Hello Andrea @ameraner 

So, I have now introduced radio buttons since push buttons didn't look "too eye catching". Also the single vs. composite bug is fixed - this was actually a logic bug in reading the list display masks for single and composite datasets.
This is how it looks now:
image

image

This is how it would look like using push buttons:
image
image
image

If you still want to have push buttons, then you could achieve that easily by replacing the lines 108ff in open_file_wizard_ui.py
```
        self.productTypeAll = QtWidgets.QRadioButton("All")
        self.productTypeSingle = QtWidgets.QRadioButton("Single")
        self.productTypeComposite = QtWidgets.QRadioButton("Composite")
```
to
```
        self.productTypeAll = QtWidgets.QPushButton("All")
        self.productTypeAll.setCheckable(True)
        self.productTypeSingle = QtWidgets.QPushButton("Single")
        self.productTypeSingle.setCheckable(True)
        self.productTypeComposite = QtWidgets.QPushButton("Composite")
        self.productTypeComposite.setCheckable(True)
```

ameraner:
https://github.com/ssec/sift/pull/466/commits/002afbb9024c3621ce539e16468446a8b7fb1425#diff-fcf4d135b50fd855e20036357f0a47cc05af9381be855fdce48925cb86191ce4 has to be reverted, I'll also probably deprecate the 3.10 tests as it's segfaulting as the 3.9 was...

ameraner (approved):
LGTM, thanks! Merging as I reverted the extra commit and the failing test is being fixed separately.

RudiLauster:
> [002afbb#diff-fcf4d135b50fd855e20036357f0a47cc05af9381be855fdce48925cb86191ce4](https://github.com/ssec/sift/commit/002afbb9024c3621ce539e16468446a8b7fb1425#diff-fcf4d135b50fd855e20036357f0a47cc05af9381be855fdce48925cb86191ce4) has to be reverted, I'll also probably deprecate the 3.10 tests as it's segfaulting as the 3.9 was...

Uhhh - I haven't noticed that this commit sneaked in. It was not planned to go into the PR as I only use it locally due to some issue with my Satpy reader setup.

Issue #411, opened 2024-05-07
Change animation button from "play" to "pause" when animation is running, and add animation speed button.


ameraner:
Currently, when an animation is running, no change in the animation buttons can be seen. This can be confusing in certain situations, e.g. when the animation speed is very low or too fast (screen update not keeping up), and the user doesn't know if the animation is running or not.

This could be improved by switching to a classic pause button || when the animation is running, or keeping the "play" button pressed until it is clicked on again.

In addition, the animation speed slider can only be changed when right-clicking the "play" button, which is not intuitive/clear to a new user. A new button or always-visible slider should be added.

Ported from EUM Gitlab issue 173.

ameraner:
Relevant piece of code: https://github.com/ssec/sift/blob/ac7d4315cdf0d34a6900bb618328b087d276f312/uwsift/__main__.py#L292-L295

PR #468, opened 2025-09-24
Refactored animation buttons and speed slider

RudiLauster:
+ animation buttons changed from unicode symbols to QT media icons, introduced pause state
+ the speed slider is now a always visible QDial with the fps shown

### DIFF
--- a/uwsift/ui/custom_widgets.py
+++ b/uwsift/ui/custom_widgets.py
@@ -1,14 +1,20 @@
 from decimal import ROUND_HALF_UP, Decimal
 
-from PyQt5.QtCore import QEvent, Qt, pyqtSignal
+from PyQt5.QtCore import QEvent, QSize, Qt, pyqtSignal
+from PyQt5.QtGui import QFont
 from PyQt5.QtWebEngineWidgets import QWebEngineView
 from PyQt5.QtWidgets import (
     QApplication,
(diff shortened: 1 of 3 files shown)


RudiLauster:
Hi Andrea @ameraner,

Is it how you imagined it? 👉 https://github.com/ssec/sift/issues/411#issuecomment-3328381871
😇

uwsift/ui/custom_widgets.py
```diff
@@ -215,3 +221,89 @@ def _round_and_count_decimals(self, value, max_decimals) -> tuple[float, int]:
...
+                # system/application font changed
+                self.update_fixed_width()
+            return super().event(event)
+
+        def update_fixed_width(self):
+            self.setFixedWidth(self.fontMetrics().boundingRect("15.0 fps").width())
+
+    _dial = None
+    _raw_minimum = 2
+    _raw_maximum = 150  # frames per 10 seconds
```
ameraner, line 309: I would set this to 100, I doubt that anyone wants to ever go above it, and this makes it a bit easier to turn the weel

ameraner (commented):
Cool, I like it! I wonder if we could make the entire icons group a bit smaller, it takes up quite a bit of vertical space now.. what do you think? And just a suggestion for the range inline.
image

